PR #7, opened 2021-01-04
[Suggestion] Add color-based search

ericcornelissen:

---

This would allow users to start their search query with a `#` to be able to search based on brand colors instead of names.

We are holding of on this feature for now as it doesn't seem very useful to us. If anyone from the community has interest in this feature we may add it.


ericcornelissen:
Keeping this unmerged for now as it doesn't seem like an incredibly useful feature. Though, given it's small size we may consider it in the future, so I'll leave it open for at least a little while.

ericcornelissen:
I will be closing this now as:

- The branch has been fallen quite far behind `master`.
- The feature is not particularly useful and there is no interest in it.
- Adding additional data to `index.html` is not a good idea at this point.

PR #30, opened 2021-01-24
Sticky controls on desktops

ericcornelissen:
This would slightly update the layout of the website on desktops (on mobile or in narrow windows the design is unchanged) by making the "controls" part (search, order, light/dark mode) stick to the top when you scroll down.

This change has two advantages:

- The controls, primarily search, is always available. Removing the need to scroll up to change the search query
- Reloading the page doesn't scroll you all the way back to the search input because it is being focused

There are no clear downsides in my opinion. It may look a little bit off, but I think that's because the website never had a sticky part.

In the long run I think we should consider going with something similar to [@PeterShaggyNoble version](https://petershaggynoble.github.io/SI-Sandbox/), where there isn't really any content before the controls and the controls are inherently sticky. However, that would require some trickery on smaller screens to accommodate all the external links - for the record, I like the side-menu approach from @PeterShaggyNoble, but it's not the only way and I think it warrants some discussion - which I don't intend to implement in this Pull Request.

https://user-images.githubusercontent.com/3742559/118378661-935ded80-b5d5-11eb-9254-818487bfd5fc.mp4


adamrusted (commented):
I've just built this on my end - and have an issue whereby the sticky menus border persists after the sticky menu returns to its default location, and can be seen poking out the top if you keep scrolling up. It does disappear pretty quickly, so it's a _very_ minor nitpick, but other than that the sticky controls work great for me.

Should we look to put the GitHub / NPM / Packagist / About links into the Sticky Header once it scrolls? Thinking providing them alongside the search box would be beneficial when scrolling down the page.

ericcornelissen:
> I've just built this on my end - and have an issue whereby the sticky menus border persists after the sticky menu returns to its default location, and can be seen poking out the top if you keep scrolling up. It does disappear pretty quickly, so it's a very minor nitpick, but other than that the sticky controls work great for me.

Which browser and OS are you using? The border should be hidden behind the controls and "header" due to it's `z-index` being lower :thinking: 


> Should we look to put the GitHub / NPM / Packagist / About links into the Sticky Header once it scrolls? Thinking providing them alongside the search box would be beneficial when scrolling down the page.

Yes I think so, e.g. something similar to @PeterShaggyNoble's design. But I think that requires some bigger design changes that I don't intent on tackling here.

adamrusted:
> Which browser and OS are you using? The border should be hidden behind the controls and "header" due to it's `z-index` being lower :thinking: 

Windows 10 and Chrome. The border is hiding behind the search bar section that _was_ sticky, but isn't hiding when under the menu bar above.

> Yes I think so, e.g. something similar to @PeterShaggyNoble's design. But I think that requires some bigger design changes that I don't intent on tackling here. 

That's fair. Let's focus on getting the sticky header in here, and iterate on it later 🤣

ericcornelissen:
> Windows 10 and Chrome. The border is hiding behind the search bar section that was sticky, but isn't hiding when under the menu bar above.

Can you share a screenshot/video?

adamrusted:
You can vaguely see the line 'reappear' in this video - unfortunately my OBS is set up at a low resolution for recording, so isn't outputting the best.

https://user-images.githubusercontent.com/1830867/118399187-e2e2fe80-b653-11eb-9bea-607b3024c1fb.mp4

PeterShaggyNoble (approved):
Looks good to me, although I'd still prefer the overall header be much more compressed, allowing people to get to the icons quicker without scrolling.

Will hold off on merging for now until the issue Adam is experiencing can be investigated further. Unfortunately, I don't have the means to run the site locally to see if I can recreate the issue.

ericcornelissen:
> Looks good to me, although I'd still prefer the overall header be much more compressed, allowing people to get to the icons quicker without scrolling.

Again I agree, but as per the Pull Request description I think there is more than one approach and we have to be careful with supporting that on smaller screens.


> Will hold off on merging for now until the issue Adam is experiencing can be investigated further.

Definitely, I will look into this as soon as possible!


> Unfortunately, I don't have the means to run the site locally to see if I can recreate the issue.

You mean you actually can't because of some limitation, or because you don't know how to? In case of the latter, [the Contributing Guielines](https://github.com/simple-icons/simple-icons-website/blob/master/CONTRIBUTING.md) should cover that. Let me know if it isn't clear.

PeterShaggyNoble:
> You mean you actually can't because of some limitation, or because you don't know how to?

I don't have Node set up; never been able to figure it out.

ericcornelissen:
I was able to reproduce the bug in Chrome @adamrusted, though a bit inconsistently. I pushed something that should prevent the bug, or at least reduce the effect. Could you retest this on your machine as well?

adamrusted (approved):
This looks good to me now, thanks for the fix @ericcornelissen! 💪

PR #44, opened 2021-02-04
Increase unit test coverage of site scripts

ericcornelissen:
Add tests for all `public/scripts` files, and improve some of those files as well. More specifically:

- `color-scheme.js`
  - Keep state in function scope (simplifies testing)
  - Check active color scheme in `selectColorScheme` (reduces duplication)
- `ordering.js`
  - Keep state in function scope (simplifies testing)
  - Check active ordering in `selectOrdering` (reduces duplication)
- `search.js`
  - Keep state in function scope (simplifies testing)
  - Make `dom-utils.js` a function parameter (simplifies testing)
  - Use `document.location` in `getQueryFromParameter` (should have been this way from the start)
  - Get ads element by ID (should have been this way from the start)

### DIFF
--- a/public/scripts/color-scheme.js
+++ b/public/scripts/color-scheme.js
@@ -10,7 +10,7 @@ const CLASS_DARK_MODE = 'dark';
 const CLASS_LIGHT_MODE = 'light';
 
 export default function initColorScheme(document, storage) {
-  let activeColorScheme = null;
+  let activeColorScheme = DEFAULT_COLOR_SCHEME;
 
   const $body = document.querySelector('body');
   const $colorSchemeDark = document.getElementById('color-scheme-dark');
@@ -24,8 +24,6 @@ export default function initColorScheme(document, storage) {
   const storedColorScheme = storage.getItem(STORAGE_KEY_COLOR_SCHEME);
   if (storedColorScheme) {
     selectColorScheme(storedColorScheme);
-  } else {
-    selectColorScheme(DEFAULT_COLOR_SCHEME);
   }
 
   $colorSchemeDark.addEventListener('click', (event) => {
--- a/public/scripts/color-scheme.js
+++ b/public/scripts/color-scheme.js
@@ -9,9 +9,9 @@ const DEFAULT_COLOR_SCHEME = COLOR_SCHEME_SYSTEM;
 const CLASS_DARK_MODE = 'dark';
 const CLASS_LIGHT_MODE = 'light';
 
-let activeColorScheme = null;
-
 export default function initColorScheme(document, storage) {
+  let activeColorScheme = null;
+
   const $body = document.querySelector('body');
   const $colorSchemeDark = document.getElementById('color-scheme-dark');
   const $colorSchemeLight = document.getElementById('color-scheme-light');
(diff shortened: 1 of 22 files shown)


public/scripts/color-scheme.js
```diff
@@ -9,9 +9,9 @@ const DEFAULT_COLOR_SCHEME = COLOR_SCHEME_SYSTEM;
 const CLASS_DARK_MODE = 'dark';
 const CLASS_LIGHT_MODE = 'light';
 
-let activeColorScheme = DEFAULT_COLOR_SCHEME;
-
 export default function initColorScheme(document, storage) {
+  let activeColorScheme = null;
```
ericcornelissen: Should this be `null` or `DEFAULT_COLOR_SCHEME`. The latter has the main advantage that nothing will be stored in local storage until the user has actually chosen a color scheme.

public/scripts/ordering.js
```diff
@@ -10,10 +10,10 @@ const CLASS_ORDER_ALPHABETICALLY = 'order-alphabetically';
 const CLASS_ORDER_BY_COLOR = 'order-by-color';
 const CLASS_ORDER_BY_RELEVANCE = 'order-by-relevance';
 
-let activeOrdering = DEFAULT_ORDERING;
-let preferredOrdering = DEFAULT_ORDERING;
-
 export default function initOrdering(document, storage) {
+  let activeOrdering = null;
+  let preferredOrdering = null;
```
ericcornelissen: Should these be `null` or `DEFAULT_ORDERING`? The latter has the main advantage that nothing will be stored in local storage until the user has actually chosen an ordering.

tests/mocks/dom.mock.js
```diff
@@ -0,0 +1,65 @@
+const PATHNAME = 'https://www.simpleicons.org';
```
ericcornelissen, line 1:
```suggestion
const PATH_NAME = 'https://www.simpleicons.org';
```

tests/mocks/local-storage.mock.js
```diff
@@ -0,0 +1,28 @@
...
+  // Utility to quickly clear the entire localStorage mock.
+  __resetAllMocks: function () {
+    STORAGE.clear();
+    this.getItem.mockReset();
+    this.getItem.mockImplementation(getItemImplementation);
+    this.setItem.mockReset();
+  },
+
+  // Utility to mock a stored value for a key
+  __storedValueFor: function (key, value) {
```
ericcornelissen:
```suggestion
  __setStoredValueFor: function (key, value) {
```

public/scripts/ordering.js
ericcornelissen: They have both been set to `DEFAULT_ORDERING` as per the advantage above. Additionally, the `DEFAULT_ORDERING` is no longer explicitly loaded [later in the file](https://github.com/simple-icons/simple-icons-website/blob/6bcf096af11224ab73b36358130e76575266b2ac/public/scripts/ordering.js#L30) either, for the same reason.

public/scripts/color-scheme.js
ericcornelissen: This has been set to `DEFAULT_COLOR_SCHEME` as per the advantage above. Additionally, the `DEFAULT_COLOR_SCHEME` is no longer explicitly loaded [later in the file](https://github.com/simple-icons/simple-icons-website/blob/6bcf096af11224ab73b36358130e76575266b2ac/public/scripts/color-scheme.js#L28) either, for the same reason.

Issue #37, opened 2020-01-17
Lazy load icons on website


ericcornelissen:
### Kind of issue 
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely: 

### This issue concerns 
  - [ ] The NPM Package
  - [ ] The Packagist Package
  - [x] The website
  - [ ] Other, namely: 

### Description

With [over a 1000 icons](https://github.com/simple-icons/simple-icons/pull/2365) I think it might be a good idea to implement the lazy loading of icons on the website. That is, when the page is first loaded we load a limited amount of icons (say 250) and when the user scrolls down we (gradually?) load more icons.

Some things to keep in mind/that should be tested thoroughly:

- How will this work with the different sorting methods (colour vs. alphabetical)
- How will this work with searching
- More...?

ericcornelissen:
> we (gradually?) load more icons.

I do think we should do it gradually. That is, I don't think we should automatically start loading all icons in the background even if the user does not interact with the page - this seems like a waste of internet connection to me.


> How will this work with the different sorting methods (colour vs. alphabetical)

I think this is one of the most difficult to tackle. Especially w.r.t. the fact that we don't know which of the two the user has selected before the page is loaded. The striaghtforward solution would be to not pre-load any SVGs, and start loading SVGs only once the ordering is known. This may work, but I think it should be tested how this affects the loading experience of the website.

Alternative suggestions are welcome.


> How will this work with searching

This should be fairly straightforward, as we can just load the SVGs that match the search query that where not previously loaded. We can easily keep list of all brand names in memory for searching (even with https://github.com/simple-icons/simple-icons/issues/4797), search in that and load the icons that are needed on demand.

A further optimization here would be to also load/show only the first X icons matching a search term, though depending on the value of X, I'm not sure if that currently has any effect.


> More 1: Supporting non-JavaScript users

Following the lead of https://github.com/simple-icons/simple-icons-website/issues/34 and https://github.com/simple-icons/simple-icons-website/issues/35, a separate page could be generated with all icons preloaded (that is, the way the website works now). This version of the website can be linked to at the end of icons being loaded by default.

ayyash:
Not sure about this lazy loading, the svgs are simple text, and the page loads fast enough, why do you want to lazy load? 

>I do think we should do it gradually. That is, I don't think we should automatically start loading all icons in the background even if the user does not interact with the page - this seems like a waste of internet connection to me.

[...]

PR #55, opened 2021-03-28
Add lazy loading using `loading="lazy"`

ericcornelissen:
Addresses #37 by replacing the inline `<svg>`s with ``s with the SVG Base-64 encoded on the `src` and `loading="lazy"`. This makes it so that the images are not loaded until they're in view. If `loading="lazy"` is not supported by the browser all icons are loaded immediately (same behaviour as current `master`).

For more details on the trade-offs of various approached we can take, see [this comment](https://github.com/simple-icons/simple-icons-website/pull/55#issuecomment-826394479).

### Preview

https://user-images.githubusercontent.com/3742559/116010501-9f324180-a61f-11eb-950f-44e41893919c.mp4

### DIFF
--- a/public/index.pug
+++ b/public/index.pug
@@ -98,8 +98,8 @@ html
 
         each icon in icons
           li(class="grid-item" style=`--order-color: ${icon.indexByColor}` data-brand=`${icon.normalizedName}`)
-            svg(class="grid-item__decoration" role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
-              use(href=`#${icon.slug}-svg`)
+            //- svg(class="grid-item__decoration" role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
+            //-   use(href=`#${icon.slug}-svg`)
             h2(class="grid-item__title") #{icon.title}
             button(
               class=`grid-item__color copy-button ${icon.light ? "contrast-light" : "contrast-dark"} ${icon.superLight ? "border-light" : ""}  ${icon.superDark ? "border-dark" : ""}`
@@ -125,9 +125,7 @@ html
               ) #{icon.license.type}
             div(class="grid-item__footer")
               button(class="grid-item__preview copy-button" title=`${icon.title} SVG` disabled)
-                svg(id=`${icon.slug}-svg` role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
-                  title #{icon.title} icon
-                  path(d=icon.path)
+                img(src=`./icons/${icon.slug}.svg` loading="lazy")
               div(class="grid-item__downloads")
                 h3 Download
                 div(class="grid-item__button-row")
--- a/public/index.pug
+++ b/public/index.pug
@@ -152,7 +152,7 @@ html
       )
         svg(class="footer-share--icon" role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
           title Twitter logo
-          use(href=`#twitter-svg`)
+          path(d=twitterIcon.path)
         span Share this on Twitter!
 
     input(id="copy-input" class="hidden" aria-hidden="true")
--- a/public/index.pug
+++ b/public/index.pug
@@ -125,7 +125,8 @@ html
               ) #{icon.license.type}
             div(class="grid-item__footer")
               button(class="grid-item__preview copy-button" title=`${icon.title} SVG` disabled)
-                img(src=`./icons/${icon.slug}.svg` loading="lazy")
+                img(src=`./icons/${icon.slug}.svg` loading="lazy" class="icon-black")
+                img(src=`./icons/${icon.slug}-white.svg` loading="lazy" class="icon-white")
               div(class="grid-item__downloads")
                 h3 Download
                 div(class="grid-item__button-row")
--- a/public/index.pug
+++ b/public/index.pug
@@ -98,8 +98,6 @@ html
 
         each icon in icons
           li(class="grid-item" style=`--order-color: ${icon.indexByColor}` data-brand=`${icon.normalizedName}`)
-            //- svg(class="grid-item__decoration" role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
-            //-   use(href=`#${icon.slug}-svg`)
             h2(class="grid-item__title") #{icon.title}
             button(
               class=`grid-item__color copy-button ${icon.light ? "contrast-light" : "contrast-dark"} ${icon.superLight ? "border-light" : ""}  ${icon.superDark ? "border-dark" : ""}`
@@ -125,7 +123,7 @@ html
               ) #{icon.license.type}
             div(class="grid-item__footer")
               button(class="grid-item__preview copy-button" title=`${icon.title} SVG` disabled)
-                img(src=`./icons/${icon.slug}.svg` loading="lazy" class="icon-preview")
+                img(src=`data:image/svg+xml;base64,${icon.base64Svg}` loading="lazy" class="icon-preview")
               div(class="grid-item__downloads")
                 h3 Download
                 div(class="grid-item__button-row")
--- a/public/index.pug
+++ b/public/index.pug
@@ -125,7 +125,7 @@ html
               ) #{icon.license.type}
             div(class="grid-item__footer")
               button(class="grid-item__preview copy-button" title=`${icon.title} SVG` disabled)
-                img(src=`data:image/svg+xml;base64,${icon.base64Svg}` loading="lazy" class="icon-preview")
+                img(src=`./icons/${icon.slug}.svg` loading="lazy" class="icon-preview")
               div(class="grid-item__downloads")
                 h3 Download
                 div(class="grid-item__button-row")
--- a/public/index.pug
+++ b/public/index.pug
@@ -125,8 +125,7 @@ html
               ) #{icon.license.type}
             div(class="grid-item__footer")
               button(class="grid-item__preview copy-button" title=`${icon.title} SVG` disabled)
-                img(src=`./icons/${icon.slug}.svg` loading="lazy" class="icon-black")
-                img(src=`./icons/${icon.slug}-white.svg` loading="lazy" class="icon-white")
+                img(src=`data:image/svg+xml;base64,${icon.base64Svg}` loading="lazy" class="icon-preview")
               div(class="grid-item__downloads")
                 h3 Download
                 div(class="grid-item__button-row")
--- a/public/scripts/copy.js
+++ b/public/scripts/copy.js
@@ -28,11 +28,10 @@ export default function initCopyButtons(document, navigator) {
       event.preventDefault();
 
       const $img = $svgButton.querySelector('img');
-      const svgUrl = $img.src;
-      const response = await fetch(svgUrl);
-      const rawSvg = await response.text();
-
-      const value = rawSvg;
+      const srcValue = $img.getAttribute('src');
+      const base64Svg = srcValue.replace('data:image/svg+xml;base64,', '');
+      console.log(base64Svg);
+      const value = atob(base64Svg);
       $svgButton.blur();
       copyValue(value);
       setCopied($svgButton);
--- a/public/scripts/copy.js
+++ b/public/scripts/copy.js
@@ -24,14 +24,14 @@ export default function initCopyButtons(document, navigator) {
 
   $svgButtons.forEach(($svgButton) => {
     $svgButton.removeAttribute('disabled');
-    $svgButton.addEventListener('click', async (event) => {
+    $svgButton.addEventListener('click', (event) => {
       event.preventDefault();
 
-      const svgUrl = $svgButton.children[0].src;
-      const response = await fetch(svgUrl);
-      const rawSvg = await response.text();
+      const $img = $svgButton.querySelector('img');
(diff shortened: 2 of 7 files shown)


public/index.pug
```diff
@@ -125,9 +125,7 @@ html
               ) #{icon.license.type}
             div(class="grid-item__footer")
               button(class="grid-item__preview copy-button" title=`${icon.title} SVG` disabled)
-                svg(id=`${icon.slug}-svg` role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
-                  title #{icon.title} icon
-                  path(d=icon.path)
+                img(src=`./icons/${icon.slug}.svg` loading="lazy")
```
service-paradis: Great!
Another downside that I did not think before: I think the `fill` property will not work anymore...
I think, the icon wont be filled in white in dark mode. Cannot try it right now though.

ericcornelissen: Good point, you're correct. Thanks for pointing it out :+1: 

I think this can be solved using plain HTML, let me investigate :thinking:

service-paradis: The possible alternative would be to use `img` with data url (svg encoded in base64). Something like this:
```
<img src="data:image/svg+xml;base64,PHN2ZyByb2xlPSJpbWciIHZpZXdCb3g9IjAgMCAyNCAyNCIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj48dGl0bGU+U2ltcGxlIEljb25zIGljb248L3RpdGxlPjxwYXRoIGQ9Ik0xMiAwQzguNjg4IDAgNiAyLjY4OCA2IDZzMi42ODggNiA2IDZjMS4wNjYgMCAyLjExNzEuMjg2MiAzLjAzNzEuODI2MiAyLjg1OCAxLjY3OCAzLjgxNjcgNS4zNTM5IDIuMTM4NyA4LjIxMjloMS42Nzk3YTcuNDkyNSA3LjQ5MjUgMCAwIDAgLjY0MjUtMy4wMjkzYy4wMDMtNC4xNDQtMy4zNTYtNy41MDQ4LTcuNDk4LTcuNTA5OC0yLjQ4NCAwLTQuNS0yLjAxNi00LjUtNC41UzkuNTE2IDEuNSAxMiAxLjVzNC41IDIuMDE2IDQuNSA0LjVIMThjMC0zLjMxMi0yLjY4OC02LTYtNnptMCAzYy0xLjU5Ny4wNC0yLjg3OTkgMS4zMjU5LTIuOTIxOSAyLjkyMTlDOS4wMzUxIDcuNTc5OSAxMC4zNDMgOC45NTcgMTIgOWMxLjU5Ny0uMDQgMi44Nzk5LTEuMzI1OSAyLjkyMTktMi45MjE5QzE0Ljk2NDkgNC40MjAxIDEzLjY1NiAzLjA0MyAxMiAzem0wIDEuNWMuODI4IDAgMS41LjY3MiAxLjUgMS41di4wMDJjMCAuODI4LS42NzIgMS41LTEuNSAxLjUtLjgzIDAtMS41LS42NzMtMS41LTEuNTAyIDAtLjgzLjY3LTEuNSAxLjUtMS41ek03LjUgMTV2MS41SDl2Nkg0LjVWMjRoMTV2LTEuNUgxNXYtNmgxLjVWMTVoLTl6bTMgMS41aDN2NmgtM3YtNnptLTYgMS40NzA3VjE4YTcuNDE4IDcuNDE4IDAgMCAwIC42NDQ1IDMuMDM5aDEuNjgzNkM2LjMwMDEgMjAuMTQ3IDYgMTkuMTEgNiAxOHYtLjAyOTNINC41WiIvPjwvc3ZnPg=="  loading="lazy" />
```
```
<img src="data:image/svg+xml;base64,PHN2ZyByb2xlPSJpbWciIHZpZXdCb3g9IjAgMCAyNCAyNCIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj48dGl0bGU+U2ltcGxlIEljb25zIGljb248L3RpdGxlPjxwYXRoIGZpbGw9IiNmZmYiIGQ9Ik0xMiAwQzguNjg4IDAgNiAyLjY4OCA2IDZzMi42ODggNiA2IDZjMS4wNjYgMCAyLjExNzEuMjg2MiAzLjAzNzEuODI2MiAyLjg1OCAxLjY3OCAzLjgxNjcgNS4zNTM5IDIuMTM4NyA4LjIxMjloMS42Nzk3YTcuNDkyNSA3LjQ5MjUgMCAwIDAgLjY0MjUtMy4wMjkzYy4wMDMtNC4xNDQtMy4zNTYtNy41MDQ4LTcuNDk4LTcuNTA5OC0yLjQ4NCAwLTQuNS0yLjAxNi00LjUtNC41UzkuNTE2IDEuNSAxMiAxLjVzNC41IDIuMDE2IDQuNSA0LjVIMThjMC0zLjMxMi0yLjY4OC02LTYtNnptMCAzYy0xLjU5Ny4wNC0yLjg3OTkgMS4zMjU5LTIuOTIxOSAyLjkyMTlDOS4wMzUxIDcuNTc5OSAxMC4zNDMgOC45NTcgMTIgOWMxLjU5Ny0uMDQgMi44Nzk5LTEuMzI1OSAyLjkyMTktMi45MjE5QzE0Ljk2NDkgNC40MjAxIDEzLjY1NiAzLjA0MyAxMiAzem0wIDEuNWMuODI4IDAgMS41LjY3MiAxLjUgMS41di4wMDJjMCAuODI4LS42NzIgMS41LTEuNSAxLjUtLjgzIDAtMS41LS42NzMtMS41LTEuNTAyIDAtLjgzLjY3LTEuNSAxLjUtMS41ek03LjUgMTV2MS41SDl2Nkg0LjVWMjRoMTV2LTEuNUgxNXYtNmgxLjVWMTVoLTl6bTMgMS41aDN2NmgtM3YtNnptLTYgMS40NzA3VjE4YTcuNDE4IDcuNDE4IDAgMCAwIC42NDQ1IDMuMDM5aDEuNjgzNkM2LjMwMDEgMjAuMTQ3IDYgMTkuMTEgNiAxOHYtLjAyOTNINC41WiIvPjwvc3ZnPg=="  loading="lazy" />
```
That would give
image

ericcornelissen: I "solved" it, though it required CSS. My initial idea was to use an approach similar to what you can find in [this article](https://flaviocopes.com/dark-mode-change-img-source/), but that doesn't work with the _forced_ dark mode and light mode :disappointed: 

If your browser supports lazy loading and JS is enabled, this solution works great. It only loads the dark mode icons once you go into dark mode - and again the cache is used whenever possible.

The downside is that if neither of these are true, the browser will end up downloading **all** SVGs **twice** when you load the website which is really, really bad...

ericcornelissen: > The possible alternative would be to use `img` with data url (svg encoded in base64). Something like this:
> 
> ```
> 
> ```
> 
> ```
> 
> ```
> 
> That would give
> image

Sorry, missed this comment. That defeats the entire goal of what we're trying to do here, which is lazy loading the SVGs. If we provide the SVG source directly in the `src` attribute of the SVG we might as well embed the SVG itself - which is what we already have on `master`.

service-paradis: > The downside is that if neither of these are true, the browser will end up downloading all SVGs twice when you load the website which is really, really bad...

Yep, that is why I was trying to find another alternative.

> That defeats the entire goal of what we're trying to do here, which is lazy loading the SVGs.

I think SVGs will still be « lazy rendered », so it might not defeats entirely the goal of this PR. But I understand that the HTML will still be large. Would surely need some test since the behavior of lazy loading inline images is not well documented.

service-paradis: I might have a compromise.
If you only add the default black svg and apply the following style on dark mode, it might do it:
```css
    filter: invert(1);
```
The support is not bad (https://caniuse.com/css-filters)

ericcornelissen: It has better support than `loading="lazy"` itself so worth a try, but indeed it's less than ideal.

ericcornelissen: > I think SVGs will still be « lazy rendered », so it might not defeats entirely the goal of this PR. But I understand that the HTML will still be large. Would surely need some test since the behavior of lazy loading inline images is not well documented.

It does indeed seem to be the case that the icons are still "lazy rendered" and it does indeed seem to still improve performance. I implemented this approach in ba07d08dbba9aa3dc138c7681e55d09af808b003.



> I might have a compromise.
> If you only add the default black svg and apply the following style on dark mode, it might do it:
> ```css
>     filter: invert(1);
> ```
> The support is not bad (https://caniuse.com/css-filters)

As this also works with external SVGs, I gave that a try as well. This fixes the problem of having to load twice as many SVGs in order to support dark mode. I implemented this approach in f8a62fc6f0871b22eba09a3b30eec3cc9677a224


---


Comparing the two approaches above, my findings are as follows:

- Both approaches are faster than the current website. A rough conservative estimate of mine is that they're both around 40% faster...
- But with external SVGs the initial load is slightly faster (if lazy loading is supported). For me the difference seems to be a little less than a second when compared to the "lazy render" approach.
- However, using external SVGs is significantly worse when lazy loading is not supported.
- Using the "lazy render" approach if lazy loading is not supported, the performance is about equal with the current website (as expected).
- If you start scrolling, external SVGs feel less responsive than "lazy rendered" SVGs (because of the time it takes to request each individual SVG).

From this it seems to me like we could definitely switch to the "lazy rendering" approach, my reasons being:

1. If lazy loading is supported, this approach slightly underperforms on the initial load but it feels more response when scrolling.
2. If lazy loading is not supported, this approach is far better than actual lazy loading.


Any @simple-icons/maintainers have an opinion on this?

ericcornelissen:
Here's a preview of what lazy loading with the current strategy looks like. Would love some opinions before we push this live.

Basically, this Pull Request can go three ways:

- Lazy loading (see below)
- Lazy rendering (see below)
- Not implementing lazy loading (see [live website](https://simple-icons.github.io/simple-icons-website/))

Comparing the two approaches where something changes, my findings are as follows:

- Both approaches are faster than the current website. A rough conservative estimate of mine is that they're both around 40% faster...
- With lazy loading the initial load time is slightly faster (provided lazy loading is supported) when compared to lazy rendering. For me, the difference seems to be a little less than a second when compared to the "lazy render" approach.
- With external SVGs the initial load time is significantly worse when lazy loading is not supported, whereas lazy rendering is roughly the same as the current approach.
- If you start scrolling, lazy loading feel less responsive than lazy rendered SVGs. I'm guessing this is because of the time it takes to request each individual SVG.

From this it seems to me like we _could_ switch to the lazy rendering approach, my reasons being:

- If lazy loading is supported, this approach slightly underperforms on the initial load but it feels more response when scrolling.
- If lazy loading is not supported, this approach is far better than actual lazy loading.

However, I say _could_ as one may consider lazy rendering not worth the trade-off. a preview of both lazy loading and lazy rendering can be seen below.


#### True lazy loading, using ``

https://user-images.githubusercontent.com/3742559/116010376-d9e7aa00-a61e-11eb-8e37-560ef806383c.mp4

#### Lazy rendering, using ``

https://user-images.githubusercontent.com/3742559/116010501-9f324180-a61f-11eb-950f-44e41893919c.mp4

adamrusted:
I think I prefer the lazy rendering personally, not sure what other @simple-icons/maintainers prefer.
I don't suppose there's a way for us to 'fade in' the cards / images as they're loaded?

ericcornelissen:
> I don't suppose there's a way for us to 'fade in' the cards / images as they're loaded?

Not as easily as lazy loading/lazy rendering, but that could be an idea to make this look a little bit nicer. Though, I do think it's crucial this animation doesn't decrease the user experience w.r.t. speed. In any case, I would prefer to leave that for a separate Pull Request.

PeterShaggyNoble:
Yeah, I think of the 2, the lazy rendering is the better option as it would seem to me that the benefits of lazy loading are somewhat negated by the additional requests necessary for each SVG. Couple it with `content-visibility:auto;` on the cards and it should improve things further.

The downside of _both_ solutions is that it makes it trickier for us to change the colour of the SVGs when switching theme. Also, I don't think (but I'm not sure) that elements in data URIs can be access by the `href` attribute of `use` nodes so we'd need to duplicate the paths of any icons we want to use in the UI.

@adamrusted, I did try fading cards in as you scroll in [my Sandbox](https://github.com/PeterShaggyNoble/SI-Sandbox/) using [ScrollReveal](https://github.com/jlmakes/scrollreveal) and it was very sluggish.

ericcornelissen:
> The downside of both solutions is that it makes it trickier for us to change the colour of the SVGs when switching theme.

As per the discussion above, this PR currently uses `filter: invert(1);` to accommodate this. [It](https://caniuse.com/css-filters) has better support [than `loading="lazy"`](https://caniuse.com/?search=loading). Admittedly, this would mean the dark mode becomes unusable for those users that ALSO don't get the benefits of lazy loading. I'd personally still be in favor of adding lazy loading/rendering as 1) the website is still usable in light mode and 2) the lazy rendering benefit is huge. 

> Also, I don't think (but I'm not sure) that elements in data URIs can be access by the href attribute of use nodes so we'd need to duplicate the paths of any icons we want to use in the UI.

You're absolutely right, I was already thinking ahead to your design update where the second SVG is removed. Even if we don't end up using your exact design, I'd definitely be in favor of just removing the second SVG if that allows us to cut the initial load time in half.

PR #65, opened 2021-05-15
Set up scheduled analysis

ericcornelissen:
Suggestion to add analysis of the repository/website through a scheduled GitHub Actions Workflow. Currently I set it to run once a month and to run:

- [A Lighthouse scan](https://github.com/marketplace/actions/lighthouse-ci-action), which will report on the "health" of the website, things such as performance, a11y, and SEO. This can be pretty useful to find improvements for the website, though our performance score is incredibly bad because of our large HTML.

  I currently configured it to just upload the results as an artifact so that anyone can look at it without having to perform the scan themselves. To be fair, anyone can do that pretty easily if they have Chrome/ium installed.
- [A CodeQL scan](https://github.com/github/codeql#readme), which runs some basic security queries on the (static) source and can report potential vulnerabilities. I think it's unlikely it will report something interesting for us, but it could still be worth it.

I'm opening this as a Pull Request because I'm not sure if everyone sees value in it. As per the above, Lighthouse is pretty easy for anyone to use and CodeQL is unlikely to find anything interesting. So if no one is interested in this we can just close it :slightly_smiling_face:

### DIFF
--- a/.github/codeql.yml
+++ b/.github/codeql.yml
@@ -1,0 +1,7 @@
+name: Simple Icons website CodeQL config
+
+queries:
+  - uses: security-extended
+
+paths:
+  - public/
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -1,0 +1,39 @@
+name: Analysis
+on:
+  push:
+  schedule:
+    - cron: '0 0 1 * *'
+
+jobs:
+  codeql:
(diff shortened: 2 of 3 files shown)


ericcornelissen:
With two :+1: from other simple-icons maintainers I will merge this in as a "trail". If it turns out it isn't useful we can remove the workflow later.

PR #73, opened 2021-05-27
Add Open Collective

### LINKED ISSUE #68: Add link to the Simple Icons Open Collective page
### Kind of issue 

- [x] Feature
- [ ] Bug
- [ ] Design
- [ ] Other, namely:

### Description

Add a link to the website that points to [the Simple Icons Open Collective page](https://opencollective.com/simple-icons).

adamrusted:
Fixes #68


ericcornelissen (changes requested):
I like the addition in the footer :+1: However, I would expect the item in the header to say "Open Collective" instead of "Donate".

Also, reconsidering, I may even go so far as to say I would prefer it only in the footer, at least with the current design of the website before the grid :thinking: What do you think @adamrusted ?

adamrusted:
Happy to go with whatever. I added it to both, but think I agree that being just in the footer is a little less "in your face" and saves top-menu space for more important links.

adamrusted:
@ericcornelissen I've removed the header link - we can investigate the potential of including something above the fold at a later date.

ericcornelissen (approved):

ericcornelissen:
Cool, thanks for the contribution @adamrusted :tada: 

> [W]e can investigate the potential of including something above the fold at a later date.

Agreed :+1:

PR #75, opened 2021-06-13
Fix Parameter of Publish Website Action

developStorm:
`single-commit` option is supposed to have a dash `-` rather than underscore `_`. [Docs](https://github.com/JamesIves/github-pages-deploy-action). 
image

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Build Website
         run: npm run build
       - name: Deploy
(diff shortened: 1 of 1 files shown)


ericcornelissen (approved):
Thank you for fixing this, @developStorm! Looks good to me :+1: 

Some context: it seems we updated this action in #70 from v3 to v4 and didn't notice the key names changed. Interesting thing is that, apart from the `SINGLE_COMMIT` option, it didn't seem to be a problem!

Issue #76, opened 2021-06-14
Redundant Building in Verify Action


developStorm:
### Kind of issue 

- [ ] Feature
- [ ] Bug
- [ ] Design
- [x] Other, namely: CI

### Description
It looks like dependencies are [reinstalled](https://github.com/simple-icons/simple-icons-website/runs/2813103103?check_suite_focus=true#step:4:1) and website is [re-built](https://github.com/simple-icons/simple-icons-website/runs/2813103103?check_suite_focus=true#step:6:8) in "Test Website" action. Therefore, it seems unnecessary to set "Build website" as a precondition for "Test Website", or maybe the whole "Build website" is an unnecessary test (already covered by other ones).

ericcornelissen:
The current design is intentional, though there is indeed some duplication. The "Build Website" job is indeed a precondition for the "Test Website" job as in order to test the website it must be possible to build the website. So, having the build job be a precondition for the test job ensures the test job will only fail because the tests fail and not because of a build error. You might be in favour of putting it in one job, I'm personally of the opinion that it's nice to be able to tell what went wrong without having to go into the logs.

We could definitely look into carrying over the website build between the two jobs, e.g. with caching or artefacts, but with the build taking just 5 seconds I'm not convinced that would be particularly useful. What do you think @developStorm ?

developStorm:
I see what you mean, I also think it may not be worthwhile to add cache for build in the existing situation. 

I am a little curious though, is there a reason you choose to let Publish Website independent of Verify? If I were to design it, I would probably hold the publish until checks pass:
![image](https://user-images.githubusercontent.com/59678453/121956815-6c9af000-cd16-11eb-940a-a09550b841c5.png)

If we go this way, then cache build phase may become actually useful -- But of course, that's a bit weird - like artificially increasing the complexity just to add caching hahaha.

developStorm:
> I'm personally of the opinion that it's nice to be able to tell what went wrong without having to go into the logs.

Having said that, if having clear status is all needed, I feel it is totally ok to run Build and Test in parallel (removing the dependency), so that the total time for waiting Action result to show out can be shortened somewhat

ericcornelissen:
> I see what you mean, I also think it may not be worthwhile to add cache for build in the existing situation.
> 
> I am a little curious though, is there a reason you choose to let Publish Website independent of Verify? If I were to design it, I would probably hold the publish until checks pass:
[...]

PR #77, opened 2021-06-15
Caching for npm

developStorm:
Implements #76

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -12,6 +12,13 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: 12.x
+      - name: Restore Cached Dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
       - name: Install dependencies
         run: npm ci
       - name: Build website
@@ -26,6 +33,13 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: 12.x
+      - name: Restore Cached Dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
       - name: Install dependencies
         run: npm ci
       - name: Run linter
@@ -41,6 +55,13 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: 12.x
+      - name: Restore Cached Dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-node-
       - name: Install dependencies
         run: npm ci
       - name: Run unit tests
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -12,7 +12,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: 12.x
-      - name: Restore Cached Dependencies
+      - name: Cache dependencies
         uses: actions/cache@v2
         with:
           path: ~/.npm
@@ -33,7 +33,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: 12.x
-      - name: Restore Cached Dependencies
+      - name: Cache dependencies
         uses: actions/cache@v2
         with:
           path: ~/.npm
@@ -55,7 +55,7 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: 12.x
-      - name: Restore Cached Dependencies
+      - name: Cache dependencies
         uses: actions/cache@v2
         with:
           path: ~/.npm
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -19,14 +19,13 @@ jobs:
           key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-node-
-      - name: Install dependencies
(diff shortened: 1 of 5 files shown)


ericcornelissen:
Thanks for the Pull Request @developStorm :ok_hand: 

I also wanted to look into caching the build website between the two jobs you originally opened the issue for. If you're up for it, feel free to include that in this Pull Request - I guess you'll need to find a cache key that is only valid for one workflow run :thinking: 

If not, let me know and I can merge this in. Just one small request, could you rename the steps to `Cache dependencies`?

developStorm:
For the build cache, just to confirm -- since we can't mix `dev` and `prod` build, the only place that this cache could apply would be between `build website` and `test website`. Is that what you are expecting? The cache key I think `github.sha` would work well as it identifies an unique commit. 

Also, we need a way to stop jest from rebuild the site automatically when cache is hit. I've tried searching but there doesn't seem to be a parameter does that (prevent rebuild/use existing build), any ideas @ericcornelissen ?

ericcornelissen:
> For the build cache, just to confirm -- since we can't mix dev and prod build, the only place that this cache could apply would be between build website and test website. Is that what you are expecting?

That is correct


> The cache key I think github.sha would work well as it identifies an unique commit.

That's a great idea! :+1: 


> Also, we need a way to stop jest from rebuild the site automatically when cache is hit. I've tried searching but there doesn't seem to be a parameter does that (prevent rebuild/use existing build).

Ah yes, I didn't think about that. The way that works is [a bit of a hack](https://github.com/simple-icons/simple-icons-website/blob/aea5cb9dfa13c08158110fad5f78894e91ccc58c/jest.config.js#L14) :sweat_smile: So, I guess I'll look into that part myself - finding a suitable solution for that hack and caching the website - does that sound good to you?

developStorm:
> The way that works is a bit of a hack

Thanks for pointing this out. I guess we could just add an ENV check here that if `process.env.action_cache` then skip build. We can then try to passing the value of [`steps.[cache step].outputs.cache-hit`](https://github.com/actions/cache#Skipping-steps-based-on-cache-hit) from Action to ENV to indicate caching status.

Or we can just check if `_site` exists, since there must not be this folder on the Action if cache isn't hit - but it may become a bit tricky when testing locally as the folder might exist due to previous builds.

*P.S. Isn't this actually using a production build? I am not seeing any command that sets `NODE_ENV=dev` 🤔.
https://github.com/simple-icons/simple-icons-website/blob/aea5cb9dfa13c08158110fad5f78894e91ccc58c/jest.config.js#L34
https://github.com/simple-icons/simple-icons-website/blob/aea5cb9dfa13c08158110fad5f78894e91ccc58c/.github/workflows/verify.yml#L18
https://github.com/simple-icons/simple-icons-website/blob/aea5cb9dfa13c08158110fad5f78894e91ccc58c/package.json#L16

ericcornelissen:
Thank you for taking the time to implement this @developStorm, though now I will need some time to review your approach and consider alternatives :slightly_smiling_face: 

> Or we can just check if _site exists, since there must not be this folder on the Action if cache isn't hit - but it may become a bit tricky when testing locally as the folder might exist due to previous builds.

That would probably be a better idea in the long run. We should be able to pull that off by comparing the last edited timestamp of `_site` and `public` but I'm not sure if that will actually work with the cache :thinking: 


> *P.S. Isn't this actually using a production build? I am not seeing any command that sets NODE_ENV=dev thinking.

Interesting. I suppose I _technically_ said "can use a dev build", not that they did. But in all seriousness, you're right and it should probably be updated to just use the dev build - at least in the CI.

developStorm:
> though now I will need some time to review your approach and consider alternatives 🙂

No problem! Just letting you know that I've now implemented caching with environment variable approach, and the verify CI is switched to use dev build now. Although it doesn't appear that the speed of integration testing has improved (・_・; [Baseline](https://github.com/simple-icons/simple-icons-website/runs/2833112148?check_suite_focus=true#step:7:1), [Share build # 1](https://github.com/simple-icons/simple-icons-website/runs/2834001268?check_suite_focus=true#step:8:1),  [Share build # 2](https://github.com/simple-icons/simple-icons-website/runs/2834020182?check_suite_focus=true#step:8:1).

ericcornelissen:
See https://github.com/developStorm/simple-icons-website/pull/1 @developStorm :slightly_smiling_face:

.github/workflows/verify.yml
```diff
@@ -12,10 +12,24 @@ jobs:
...
+        with:
+          path: .cache/webpack
+          key: ${{ runner.os }}-build-${{ hashFiles('**/package-lock.json') }}
       - name: Install dependencies
         run: npm ci
       - name: Build website
-        run: npm run build
+        run: npm run build:dev
+        env:
+          TEST_ENV: true # Slice icon list
```
ericcornelissen: Didn't notice this before, I don't feel like this is necessary here so I prefer to remove it - that's one less place to change it if this is changed in the future.

```suggestion
        run: npm run build:dev
```

ericcornelissen (changes requested):
One more request for change below that I missed in my Pull Request against your changes.

Also, now that the build is cached and shared between CI runs I'm starting to think that maybe we should revert back to production builds in the Build and Test jobs. While in theory both builds should work, this _may_ not be the case due to a bug. And, ultimately it's the production build that we want to ensure works. On the other hand testing with a development build locally is nicer. So, I haven't quite settled on a decision in this regard so feel free to leave this as is and I'll probably merge.

.github/workflows/verify.yml
developStorm: I think it would be better to use prod build, as we always "test" to make sure it is available in a production environment, not a test environment. Using prod build also doesn't seem to result in less debugging information, so it won't affect local debug adversely. I'll revert related changes.

ericcornelissen (approved):
Looking great 👍 Thanks for the work and discussion @developStorm!


> I think it would be better to use prod build, as we always "test" to make sure it is available in a production environment, not a test environment. Using prod build also doesn't seem to result in less debugging information, so it won't affect local debug adversely. I'll revert related changes.

I'm happy with this decision for now. We can always play around with the configuration to speed up testing locally if needed (e.g. an option to select between dev and prod build when running tests).

PR #80, opened 2021-06-24
Improve CI caching

ericcornelissen:
- [x] Add Jest caching (ref. https://github.com/simple-icons/simple-icons/pull/5912)
- [x] Add `restore-keys` for build cache

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -62,10 +62,6 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: 12.x
-      - name: Get Date
-        id: get-date
-        shell: bash
-        run: echo "::set-output name=date::$(/bin/date -u "+%Y%m%d")"
       - name: Cache dependencies
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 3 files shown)


ericcornelissen:
@developStorm, any thoughts no this? :slightly_smiling_face:

.github/workflows/verify.yml
```diff
@@ -72,6 +78,16 @@ jobs:
         with:
           path: .cache/webpack
           key: ${{ runner.os }}-build-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-
+      - name: Cache Jest
+        uses: actions/cache@v2
+        with:
+          path: .cache/jest
+          key: ${{ runner.os }}-jest-${{ hashFiles('package-lock.json') }}-${{ steps.get-date.outputs.date }}
```
developStorm: Why do we use the date as part of the cache key here? We can probably use [`${{ github.run_number }}`](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) here if this is only for sorting

developStorm (commented):
Looks good to me! Just have a small question

.github/workflows/verify.yml
ericcornelissen: The idea of using the date was to keep the cache only valid for a limited amount of time - otherwise the cache is only replaced if the project dependencies are updated, even if the tests or source is updated. Judging from [the `restore-keys` docs](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) I think your suggestion would have approximately the same effect, the only real difference being the number of caches that will be created.

So, I think we should just go ahead and use `${{ github.run_number }}` instead. Thanks for the suggestion @developStorm!

developStorm (approved):

PR #83, opened 2021-06-28
Improve HTML

ericcornelissen:
This Pull Request aims to improve the HTML of the website. These improvements are based on [this checklist](https://github.com/thedaviddias/Front-End-Checklist#best-practices) by [David Dias](https://github.com/thedaviddias) and [this checklist](https://github.com/joshbuchea/HEAD) by [Josh Buchea](https://joshbuchea.com/).

This is currently an ongoing process, other suggestions regarding the HTML are welcome 🙂 

_**Note**: commit 2dbfbcf141402e31f484524c2da53c5006bfb74b should have been part of this Pull Request, it covers [the Head section of the checklist](https://github.com/thedaviddias/Front-End-Checklist#head)._

### Enhancements

- Add `role="button"` to button-like `<a>` elements.
- Link [the license](https://github.com/simple-icons/simple-icons-website/blob/cc41ddcf3ce318611982daaabd8961da13f33a52/LICENSE.md) in the `<head>` through a `rel="license"` `<link>`. (somewhat related to https://github.com/simple-icons/simple-icons/pull/1154)

### Bugfixes

- Fix incorrect usage of `aria-hidden`, thanks @service-paradis (more about [`aria-hidden="true"`](https://stackoverflow.com/a/55061234)).
- Add missing `aria-labelledby` to control buttons, thanks @service-paradis.

### Performance

- Remove unnecessary `rel="noopener"` if `target="_blank"` is not used (more about [`rel="noopener"`](https://mathiasbynens.github.io/rel-noopener/)).
- Remove unnecessary role attribute on `<header>` (based on [w3 HTML checker scan](https://user-images.githubusercontent.com/3742559/123710289-c5b66800-d86e-11eb-8ccc-888494af9d99.png)).
- Remove unnecessary role attribute on `<main>` (based on [w3 HTML checker scan](https://user-images.githubusercontent.com/3742559/123710289-c5b66800-d86e-11eb-8ccc-888494af9d99.png)).
- Remove unnecessary role attribute on `<footer> (based on [w3 HTML checker scan](https://user-images.githubusercontent.com/3742559/123710289-c5b66800-d86e-11eb-8ccc-888494af9d99.png)).

---

### Tried/Investigated but not included

- Using HTML meta tags for cache control
  - **Idea**: Our content generally changes only once a week (when the icons are updated) so we might benefit from long-term caching.
  - **Problem**: We can only configure caches through HTML meta tags as we don't control the web server. Howeverm HTTP headers take precedence over those, and GitHub defines the relevant headers automatically.


public/index.pug
```diff
@@ -25,24 +25,24 @@ html(lang="en-US")
...
         li(class="header__list-item")
-          a(href="https://www.jsdelivr.com/package/npm/simple-icons" rel="noopener") JSDelivr
+          a(href="https://www.jsdelivr.com/package/npm/simple-icons") JSDelivr
         li(class="header__list-item")
-          a(href="https://unpkg.com/browse/simple-icons/" rel="noopener") Unpkg
+          a(href="https://unpkg.com/browse/simple-icons/") Unpkg
       h1(class="header__title") Simple Icons
       p(class="header__description") #{iconCount} Free #[abbr(title="Scalable Vector Graphic") SVG] icons for popular brands
 
     div(class="sticky-header-shadow" aria-hidden)
```
service-paradis: While improving the html, it would be good to change this for `aria-hidden="true"`
```suggestion
    div(class="sticky-header-shadow" aria-hidden="true")
```
Right now, it renders `aria-hidden=""` which is an invalid value

service-paradis (approved):
LGTM 💯 

I would add another suggestion. For accessibility purpose, there are some SVGs without accessible text (buttons used to sort and to change scheme).

It would be a good idea to add `aria-labelledby`. Ex.

```pug
button(id="order-alpha" class="control__button" title="Sort alphabetically" disabled)
  svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" aria-labelledby="order-alpha")
```

public/index.pug
ericcornelissen: Thanks for the suggestion, I always assumed `aria-hidden`/`aria-hidden=""` has the same effect as `aria-hidden="true"` but indeed it is not. I'll fix all instances of this across the website :+1:

ericcornelissen:
> I would add another suggestion. For accessibility purpose, there are some SVGs without accessible text (buttons used to sort and to change scheme).
> 
> It would be a good idea to add `aria-labelledby`. Ex.

Good catch, fixed this in 1ee1c6cc07db437d2ec9bb66ea4d93caeac3d0e9 :+1:

PR #85, opened 2021-07-04
Update search input

ericcornelissen:
Part of the effort started in #83, but separated as I expect this will take some more discussion than that Pull Request requires.

This changes the input type of the search bar from `"text"` to `"search"`. This makes semantic sense and allows us to easily hide the keyboard on mobile devices when the users presses the "search" button on their keyboard by adding an event listener for the `"change"` event.

However, there's a catch: some browsers automatically add a button to search inputs to clear it (notably Chrome), while other don't (notably Firefox, desktop 78.0.2; mobile 89.1.1). Does anyone has suggestions or opinions for solving this problem (@simple-icons/maintainers)? Can we remove/hide it in browsers like Chrome? Or should we just drop it from other browsers?

### DIFF
--- a/public/index.pug
+++ b/public/index.pug
@@ -52,7 +52,7 @@ html(lang="en-US")
             input(
               id="search-input"
               class="control__input"
-              type="text"
+              type="search"
               placeholder="Search by brand..."
               disabled
             )
--- a/public/scripts/search.js
+++ b/public/scripts/search.js
@@ -61,7 +61,7 @@ export default function initSearch(history, document, ordering, domUtils) {
       search(value);
     }),
   );
-  $searchInput.addEventListener('search', () => {
+  $searchInput.addEventListener('change', () => {
     $searchInput.blur();
   });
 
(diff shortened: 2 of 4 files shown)


runxel:
Various [sources](https://stackoverflow.com/questions/9421551/how-do-i-remove-all-default-webkit-search-field-styling) suggest that a `-webkit-appearance:none;` will work wonder.
I think it's easier to remove those unspecified elements.

ericcornelissen:
> Various [sources](https://stackoverflow.com/questions/9421551/how-do-i-remove-all-default-webkit-search-field-styling) suggest that a `-webkit-appearance:none;` will work wonder.
> I think it's easier to remove those unspecified elements.

Thanks for the pointers @runxel! works like a charm for me. Any chance any other @simple-icons/maintainers can test this out on their end as well?

jorgeamadosoria:
Looks good to me.
I tested it on desktop and mobile for Chrome, Firefox, Safari and Brave.
It works just fine, no buttons are shown as expected. 
The UX is the same as before, but now the code makes more semantic sense.
It also works as expected in Mobile Chrome, at least on my Android.
I would say this is ready for merge

PR #90, opened 2021-07-13
added required CSS reset

### LINKED ISSUE #89: Add a CSS reset
### Kind of issue 

- [ ] Feature
- [ ] Bug
- [ ] Design
- [x] Other, namely: refactor

### Description

We should consider adding a CSS reset (e.g. [Reset CSS](https://meyerweb.com/eric/tools/css/reset/) by _Eric A. Meyer_) to the website to improve consistency across browsers.

To anyone looking to work on this, please be sure to remove duplicated styles from existing CSS rules.

priyansh32:
#89 
added CSS reset and removed duplicated styles (padding and outline)

### DIFF
--- a/public/stylesheet.css
+++ b/public/stylesheet.css
@@ -1,4 +1,6 @@
-* {
+*,
+*::before,
+*::after {
   border: 0;
   box-sizing: border-box;
   font: inherit;
@@ -194,7 +196,7 @@ body {
   display: flex;
   flex-direction: column;
   line-height: 1.4;
-  min-height: 100%;
+  min-height: 100vh;
 }
 
 button {
--- a/public/stylesheet.css
+++ b/public/stylesheet.css
@@ -1,14 +1,14 @@
 * {
   box-sizing: border-box;
-  margin: 0;
-  padding: 0;
-  outline: 0;
   font: inherit;
+  margin: 0;
+  outline: 0;
+  padding: 0;
   vertical-align: baseline;
 }
 
 :root {
-  font-size: 100%;
+  font-size: 100%; 
   height: 100%;
   min-width: 280px;
 
@@ -191,9 +191,9 @@ body.dark .icon-preview {
 body {
   background-color: var(--color-background);
   color: var(--color-text-default);
-  line-height: 1;
   display: flex;
   flex-direction: column;
+  line-height: 1;
   min-height: 100%;
 }
 
--- a/public/stylesheet.css
+++ b/public/stylesheet.css
@@ -8,7 +8,7 @@
 }
 
 :root {
-  font-size: 100%; 
+  font-size: 100%;
   height: 100%;
   min-width: 280px;
 
--- a/public/stylesheet.css
+++ b/public/stylesheet.css
@@ -2,6 +2,9 @@
   box-sizing: border-box;
   margin: 0;
   padding: 0;
+  outline: 0;
+  font: inherit;
+  vertical-align: baseline;
 }
 
 :root {
@@ -188,6 +191,7 @@ body.dark .icon-preview {
 body {
   background-color: var(--color-background);
   color: var(--color-text-default);
+  line-height: 1;
   display: flex;
   flex-direction: column;
   min-height: 100%;
@@ -197,7 +201,11 @@ button {
   background: none;
   border: none;
   cursor: pointer;
-  outline: 0;
+}
+
+ol,
+ul {
+  list-style: none;
 }
 
 a {
@@ -245,7 +253,6 @@ a {
   flex-wrap: wrap;
   float: right;
   justify-content: flex-end;
-  list-style: none;
   margin: 0.2rem 0 0;
 }
 @media (max-width: 640px) {
@@ -363,7 +370,6 @@ body.no-js .control-field__title {
 
 .control__input {
   font-family: var(--font-family-stylized);
-  outline: 0;
   padding: 0.4rem 2rem 0.4rem 0.8rem;
   width: 100%;
 }
@@ -467,7 +473,6 @@ body:not(.dark):not(.light) #color-scheme-system:not(:disabled) {
   --grid-width: 13rem;
 
   display: grid;
-  list-style: none;
   gap: 0.75rem;
   grid-template-columns: repeat(auto-fill, minmax(var(--grid-width), 1fr));
   margin: 0 0 1.5rem;
@@ -604,7 +609,6 @@ body:not(.dark):not(.light) #color-scheme-system:not(:disabled) {
   font-family: var(--font-family-stylized);
   font-size: 0.75rem;
   font-weight: 600;
-  outline: none;
   padding: 8px;
   text-align: center;
 }
@@ -622,7 +626,6 @@ body:not(.dark):not(.light) #color-scheme-system:not(:disabled) {
   box-shadow: none;
   grid-column: -3 / -1;
   grid-row: span 2;
-  padding: 0;
   text-align: center;
 }
 @media (max-width: 1280px) {
--- a/public/stylesheet.css
+++ b/public/stylesheet.css
@@ -193,13 +193,12 @@ body {
   color: var(--color-text-default);
   display: flex;
   flex-direction: column;
-  line-height: 1;
+  line-height: 1.4;
   min-height: 100%;
 }
 
 button {
   background: none;
-  border: none;
   cursor: pointer;
   outline: none;
 }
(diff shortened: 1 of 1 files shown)


ericcornelissen:
Thank you for taking your time to work on this @priyansh32! :tada: 

To help me review your Pull Request, could you answer the following questions and also provide additional details that you think might be relevant.

- Which CSS reset did you use as a starting point, if any?
- Did you add all the properties that are in that reset? If not, which ones did you omit and why?
- Did you add all the selectors that are in that reset? If not, which ones did you omit and why?

Additionally, one minor request for changes is to alphabetically order the properties in the rulesets of `*` and `body`.

priyansh32:
Hey @ericcornelissen 

I referred to [https://meyerweb.com/eric/tools/css/reset/](https://meyerweb.com/eric/tools/css/reset/) for the reset.
I added all properties in the reset to the global selector (*) and omitted the below part as it is no more required in all modern browsers
```CSS
/* HTML5 display-role reset for older browsers */
article, aside, details, figcaption, figure, 
footer, header, hgroup, menu, nav, section {
	display: block;
}
```
and I omitted this as these elements are not used in the website
```CSS 
blockquote, q {
	quotes: none;
}
blockquote:before, blockquote:after,
q:before, q:after {
	content: '';
	content: none;
}
table {
	border-collapse: collapse;
	border-spacing: 0;
}
```

I have arranged properties in alphabetical order and committed changes.
I am new to Open Source and this was my first contribution to an Open Source project and I am looking forward to making more contributions to [simpleicons.org](https://simpleicons.org/) and other projects.

public/stylesheet.css
```diff
@@ -1,11 +1,14 @@
 * {
   box-sizing: border-box;
+  font: inherit;
   margin: 0;
+  outline: 0;
   padding: 0;
+  vertical-align: baseline;
 }
```
ericcornelissen: Meyer's reset includes `border: 0`, why didn't you include it here?

public/stylesheet.css
```diff
@@ -1,11 +1,14 @@
 * {
   box-sizing: border-box;
+  font: inherit;
   margin: 0;
+  outline: 0;
```
ericcornelissen: I understand your motivation to move `outline: 0` here, but in general it's not a good practice to remove `outline` from all (or any) elements - [read more about it on this website](http://www.outlinenone.com/). Could you revert the changes related to the `outline` property, and possibly setting them all to `none` instead of `0` ([ref](https://stackoverflow.com/questions/35648667/outline-none-vs-outline-0)).

The motivation for having this rule on some elements is that they have a clear focus state (though this is topic that's for debate if any a11y experts disagre).

public/stylesheet.css
```diff
@@ -1,11 +1,14 @@
...
   padding: 0;
+  vertical-align: baseline;
 }
 
 :root {
-  font-size: 100%;
+  font-size: 100%;
```
ericcornelissen: Did you try moving this to the `*` ruleset, similar to Meyer's reset?

ericcornelissen (changes requested):
I have a couple more points for changes below :slightly_smiling_face: 

> I referred to https://meyerweb.com/eric/tools/css/reset/ for the reset.

Thanks for clarifying

> I added all properties in the reset to the global selector (*) and omitted the below part as it is no more required in all modern browsers
> 
> ```css
> /* HTML5 display-role reset for older browsers */
> article, aside, details, figcaption, figure, 
> footer, header, hgroup, menu, nav, section {
> 	display: block;
> }
> ```
> 
> and I omitted this as these elements are not used in the website
> 
> ```css
> blockquote, q {
> 	quotes: none;
> }
> blockquote:before, blockquote:after,
> q:before, q:after {
> 	content: '';
> 	content: none;
> }
> table {
> 	border-collapse: collapse;
> 	border-spacing: 0;
> }
> ```

Both of those decisions seems sensible to me, nice!


> I have arranged properties in alphabetical order and committed changes.

Thanks for that :slightly_smiling_face: 


> I am new to Open Source and this was my first contribution to an Open Source project and I am looking forward to making more contributions to [simpleicons.org](https://simpleicons.org/) and other projects.

Cool, we always like fist time contributions :smile: No problem not providing these details up front, but keep it in mind for future contributions.

public/stylesheet.css
priyansh32: I would do that, I chose not to modify it as wasn't affecting any styles

public/stylesheet.css
priyansh32: thanks for that, I would make this changes.

public/stylesheet.css
priyansh32: I have added it, I  mistakenly omitted it earlier😬

priyansh32:
committed the requested changes.

public/stylesheet.css
```diff
@@ -190,14 +193,20 @@ body {
   color: var(--color-text-default);
   display: flex;
   flex-direction: column;
+  line-height: 1;
```
ericcornelissen: This rule causes the text to be a bit close together compared to the current design (comparison below). I suggest we deviate from Mayer's reset and, based on [the MDN line-height](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#values) ("`normal`: [...] use a default value of roughly `1.2`") and some local testing, set this to `1.4`. What do you think @priyansh32?

```suggestion
  line-height: 1.4;
```

### Example comparison

| This Pull Request | Live site | `line-height: 1.4` |
| --- | --- | --- |
| image | image | image |

public/stylesheet.css
```diff
@@ -1,11 +1,14 @@
 * {
+  border: 0;
```
ericcornelissen, line 4: There's another `border: 0;` in the stylesheet, can you remove that one?

ericcornelissen (changes requested):
Nice @priyansh32, I think this Pull Request is almost finished. I did a full review now and I only have two minor requests for changes left.

~~For anyone interested, here's a comparison of the website's design between ca38c92991c5681d565088db0d595342d2a1e879 (PR) and 48119c0cc127f6c6e7f7d3b50d20a16de5466a6f (`master`):~~

public/stylesheet.css
priyansh32: Done, there was one more `border:none;` removed that too

public/stylesheet.css
priyansh32: Yes,1.4 is good, added

priyansh32: The page is too long, I wonder if anybody would ever scroll down to footer, I think the content needs to be paginated.

public/stylesheet.css
ericcornelissen: Missed that myself, thanks for checking and removing :ok_hand:

public/stylesheet.css
ericcornelissen: > Yes,1.4 is good, added

Thanks!


> The page is too long, I wonder if anybody would ever scroll down to footer, I think the content needs to be paginated.

Not sure what you mean by "too long", but pagination is something we could consider. However, that's outside the scope of this Pull Request, feel free to open an issue for it though :slightly_smiling_face: 

The main challenge with pagination will be that we don't have a back-end and that we still need to support search over all icons - not impossible just something to consider. Also, I'm personally in favor of having one long page, though improving initial load time is a concern that pagination would solve - though I'd personally prefer "full" lazy loading.

ericcornelissen (approved):
Looking good @priyansh32, nice work :ok_hand: I'll leave this open for a while to give another @simple-icons/maintainers time to review and compare (see preview below).

If there's no response in a few days I will just merge this in - there's always later to improve :smile: In the meanwhile, could you merge `master` into this Pull Request @priyansh32?

| Live website | Pull Request website |
| --- | --- |
| live website preview | pr website preview |

runxel (commented):
Thanks for the PR @priyansh32!
Looks good from my pov.
However I would like to suggest adding another one or two things:
Based on the "[modern CSS reset](https://github.com/andy-piccalilli/modern-css-reset/blob/master/dist/reset.css)"

```css
/* Box sizing rules */
*,
*::before,
*::after {
  box-sizing: border-box;
}

body {
  min-height: 100vh;
}
```

At the moment `body` is set to be 100%, but this doesn't work if there is no height set at the parent (which would be `html`). Using a non-percentage unit is better, I think.

ericcornelissen (approved):
Nice work @priyansh32! With the review points by @runxel resolved I think this Pull Request is ready to be merged :ok_hand: 

Congratulations on your first open source contribution :confetti_ball:

PR #92, opened 2021-07-29
Use Node.js 16

LitoMore:
~~Consistent with the Node.js version in Docker.~~

### DIFF
--- a/.github/workflows/analysis.yml
+++ b/.github/workflows/analysis.yml
@@ -23,10 +23,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Use Node.js 12.x
+      - name: Use Node.js 16
         uses: actions/setup-node@v2
         with:
-          node-version: 12.x
+          node-version: 16
       - name: Install dependencies
         run: npm ci
       - name: Build website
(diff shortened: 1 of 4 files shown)


ericcornelissen:
Any particular reason you closed this @LitoMore? Depending on the outcome of https://github.com/simple-icons/simple-icons/pull/6179 I do think we want to merge this Pull Request.

LitoMore:
You have solved my concerns!

Let's go for it!

LitoMore:
@ericcornelissen Do we need to drop lockfile for this repo?

ericcornelissen:
I think that the dependencies in this project (as well as [font](https://github.com/simple-icons/simple-icons-font), [release-action](https://github.com/simple-icons/release-action), and probably [pdfs](https://github.com/simple-icons/simple-icons-pdf)) are a bit more "advanced" than the dependencies of the main project and I don't think it's a good idea to just drop the `package-lock.json` here because of that. Not to mention that the "problem" is much less severe here due to there being less activity.

That's not to say we shouldn't do anything in this (or any of the other projects). If you don't mind, should we use #103 and #104 instead to resolve the discussion?

LitoMore:
Agree.

PR #93, opened 2021-07-29
Fix access permission of husky files

LitoMore:
This hint appeared when I try to run `git commit` or `git push`.

```
hint: The '.husky/pre-commit' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
```

I think we should make those files executable as [simple-icons-font](https://github.com/simple-icons/simple-icons-font/blob/develop/.husky/pre-commit) did.


ericcornelissen (approved):
Thanks for fixing this @LitoMore :confetti_ball:

Issue #58, opened 2021-05-08
Consider design for link to DISCLAIMER.md


ericcornelissen:
### Kind of issue 

- [ ] Feature
- [ ] Bug
- [x] Design
- [ ] Other, namely: _requirement_

### Description

~~Add a link to the `DISCLAIMER.md` of the project - depends on https://github.com/simple-icons/simple-icons/pull/4912.~~

In #74 we were not necessarily happy about the design aspect of the link to the disclaimer, but we merged it in in the interest of having some link while we figure out the design. This ticket now serves to keep this topic alive and potentially discuss the design of the link to the disclaimer.

jorgeamadosoria:
This issue is included in the changes discussed in #39 .
I think we can close it and continue the discussion there.

ericcornelissen:
Not sure if I agree. This issue exists because, at the time that #74 was merged, we thought that it would be better to put the disclaimer somewhere else. However, we thought this would require more work and considered this link to be of a high enough priority to be included immediately. #39 merely proposes to change the header links from text to icons. Even if we do decide to keep the disclaimer link in the header, what icon would we use for it in the context of #39? Or would we keep it as text?

jorgeamadosoria:
there is a proposal in #39 for that. @PeterShaggyNoble uses a gavel for the disclaimer. Not a bad idea, maybe could be improved.
It could be an information icon for an about page with a dropdown that includes the disclaimer.
It could be many things, all of which will be by necessity discussed in #39.
We can keep the ticket open, but whatever gets decided in #39 will affect this one, to the point of closing it, even.

ericcornelissen:
> there is a proposal in #39 for that. PeterShaggyNoble uses a gavel for the disclaimer. Not a bad idea, maybe could be improved.

I think I know which image you're referring to, but it's not in #39 nor in #74 (and I don't know where it is). While I don't disagree the gavel is a decent choice, I do wonder if it's clear enough for the disclaimer. To me, the link to the disclaimer should be very clear so that no one - whether you're looking for it or not - can misunderstand it. I think a gavel icon can easily be interpreted as being something else (e.g. a link to a license or terms of service).


> It could be many things, all of which will be by necessity discussed in #39.
> We can keep the ticket open, but whatever gets decided in #39 will affect this one, to the point of closing it, even.

[...]

Issue #49, opened 2021-02-22
Add link pointing to simple-icons-website repository


mondeja:
### Kind of issue

- [x] Feature
- [ ] Bug
- [ ] Design
- [ ] Other, namely:

### Description
I miss a link that takes you to the web repository (this repository), as long as it does not confuse with the main repository. Do you see it too meta? Maybe it could be "hidden" at the end, like the Twitter share button ... :thinking:

ericcornelissen:
I'm not sure what it is you're looking for. Why should there be a link to this repository on the website? Do you want people to show where they can find the source if they're interested (if so, the footer seems right)? Do you want to make sure people open an issue for the website in the correct repository (if so, I think we should just link to this repo the same way [this repo links to the main repo](https://github.com/simple-icons/simple-icons-website/issues/new/choose))? Anything else? 

To clarify, I'm not against it, just not sure why it is needed 🙂 

On a related note, I do think we can consider updating the link "GitHub" to point to `https://github.com/simple-icons` instead of `https://github.com/simple-icons/simple-icons` (relates to https://github.com/simple-icons/simple-icons-website/issues/27).

mondeja:
> Why should there be a link to this repository on the website? Do you want people to show where they can find the source if they're interested (if so, the footer seems right)?

I don't say that "should", only seems to me that a buttton like "fork me on Github" it could help to link this repository from the website, perhaps for an user who wants to fix something at any given time, or just want to see how something is done, but obviously less intrusive that the typical "fork me on Github" in the corner.

> On a related note, I do think we can consider updating the link "GitHub" to point to `https://github.com/simple-icons` instead of `https://github.com/simple-icons/simple-icons`...

I agree :+1:

ericcornelissen:
Apologies if my response seemed negative, I was genuinely asking why we would add it and listing reasons I could come up with; I just don't think we should add it just because "every" open source website has the _"Fork me on GitHub"_ banner or something similar :slightly_smiling_face: 

If you want to make it easier to find the source of the website to either look at it and to submit issues or fixes - beyond the link to GitHub in the header -then I'm definitely open to that. And I agree that it should not be too intrusive, as it isn't that important I would say. Maybe something in the footer with the right text so it is easily found with <kbd>Ctrl</kbd>+<kbd>F</kbd>?

mondeja:
[...]

Issue #39, opened 2021-01-26
[Suggestion] Replace header link text by logos


ericcornelissen:
**:information_source: If you agree with this design idea, please leave comment or add a :+1:**

---

### Kind of issue 

- [ ] Feature
- [ ] Bug
- [x] Design
- [ ] Other, namely:

### Description

> Related issues: https://github.com/simple-icons/simple-icons-website/issues/27

Currently, the website header contains three links to external platforms (and an about link), namely: GitHub, NPM, and Packagist. These links are in text, but it is common for such links to be icons - and so did many redesign mockups in https://github.com/simple-icons/simple-icons/issues/980. 

So the question is, should we replace these text links by icon links? I think I would be in favor of this as I think it looks better, it is arguably easier to parse an image than it is text. On the other hand, it is usually recommended to use text for accessibility - though workarounds for that exist.

Alternatively, we could also choose to have both text and icons, hiding the text in some kind of [compact mode](https://github.com/simple-icons/simple-icons-website/issues/17).

PR #96, opened 2021-08-15
added report icon to the main menu

### LINKED ISSUE #95: Add link to report outdated icons
### Kind of issue 

- [X] Feature
- [ ] Bug
- [ ] Design
- [ ] Other, namely:

We should make it easier for users to report outdated icons in an effort to keep our collection up-to-date. Based on discussion (see below) we think the link should basically just point to [the Icon update template](https://github.com/simple-icons/simple-icons/issues/new?assignees=&labels=icon+outdated&template=icon_update.md) as (a) this reduces the chances (prevents?) of someone creating many issues quickly in an automated way, and (b) while prefilling details such as the brand name is _nice_, it doesn't add to much value to the UX while adding a lot of HTML (at least one element per icon = at least 1900 elements).

jorgeamadosoria:
Added a "Report Icon" link to the menu to open the "Icon Update" template and simplify icon reporting in the main repository.


ericcornelissen (commented):
I don't think "Report icon" is clear enough as it could also mean something like "report an icon that should be removed".

Ultimately, I think we need some kind of redesign to include this (as it will probably need a longer string for clarity), adding it to the header links isn't going to cut it anymore - see also #39, #49, and #58 - do you have any ideas?

jorgeamadosoria:
- #49  I don't care one way or the other. If I was forced to choose, I wouldn't add it. A link to the website repo is not that relevant. The product is the icon library, the website is merely a showcase.
- #58 is superseded by #39. We could close it, even.
- #39 is good, and we should do it. It would also help add the third parties links (there is a ticket for that) and any other links necessary in the future

ericcornelissen:
Could you leave comments with you thoughts for each of these on those particular issues?

The larger point, that brings those issues and this Pull Request together, is that (1) the header needs some work to begin with and (2) throwing more and more links in the header isn't sustainable; the link to the disclaimer (arguably) shouldn't even be there (we put it there for now because it's a very important link). By "do you have any ideas" I meant do you have any ideas on how we could redesign this in a more sustainable way? (_apologies for not being very clear the first time around_ :sweat_smile:)

jorgeamadosoria:
> Could you leave comments with you thoughts for each of these on those particular issues?
> 
> The larger point, that brings those issues and this Pull Request together, is that (1) the header needs some work to begin with and (2) throwing more and more links in the header isn't sustainable; the link to the disclaimer (arguably) shouldn't even be there (we put it there for now because it's a very important link). By "do you have any ideas" I meant do you have any ideas on how we could redesign this in a more sustainable way? (_apologies for not being very clear the first time around_ 😅)

I like this proposal: https://petershaggynoble.github.io/SI-Sandbox/preview/
It needs an about icon to handle more than one Simple Icons link; a Third party icon to contain a dropdown menu for all of the modules, a report menu if you want to give separate links for reporting different things, etc.
Alternatively, just leaving it like that and add an icon for outdated icon reporting.

jorgeamadosoria:
I've added changes to this PR to conform to the changes in #99 . The icon I chose is a placeholder, if anyone has a different option, it's very likely to be better than this one

ericcornelissen:
I personally don't think the chosen icon (see 
[preview](https://user-images.githubusercontent.com/3742559/132106550-00f6fd37-d0b1-4aab-a7f4-60fbb87d5282.png)) clearly conveys where the link is taking you - or even that it is a link at all (though it being next to the other links should make that somewhat clear, it could also just be interpreted as "there's a warning and we choose to put it here").

An alert symbol could mean so many things and, in my experience, is rarely used by itself as a button/link. If I visited the website as a new user, I would have no idea what this button did and I would definitely not expect it to take me to a new GitHub issue page.

I strongly feel like we should "add this because it's nice to have a convenient link on the website to report outdated icons" if there's even the slightest chance (a not-significant group of) people will get confused.

Also, the link should be `https://github.com/simple-icons/simple-icons/issues/new?assignees=&labels=icon+outdated&template=icon_update.yml&title=Update%3A+`.

ericcornelissen:
I think this should definitely not be merged with [the current icon](https://user-images.githubusercontent.com/3742559/132106893-e4394c24-0824-4aeb-b75a-ec883ed875d7.png). The current in no way communicates that it will let the user report an outdated icon. From my perspective it just look like an interactive warning, which if clicked will tell you what the warning is.


I have no suggestion for a better icon. I strongly believe we're better off not adding anything in this regard than adding something that is confusing. I think the only suitable solution is to do in text form. Also, we could perhaps start by adding a link like this to the README of the main project as a trail, to see if having something like this actually increases the number of outdated icons reported.

Also the link is incorrect, it should be: `https://github.com/simple-icons/simple-icons/issues/new?labels=icon+outdated&template=icon_update.yml&title=Update%3A+`

jorgeamadosoria:
> I think this should definitely not be merged with [the current icon](https://user-images.githubusercontent.com/3742559/132106893-e4394c24-0824-4aeb-b75a-ec883ed875d7.png). The current in no way communicates that it will let the user report an outdated icon. From my perspective it just look like an interactive warning, which if clicked will tell you what the warning is.
> 
> I have no suggestion for a better icon. I strongly believe we're better off not adding anything in this regard than adding something that is confusing. I think the only suitable solution is to do in text form. Also, we could perhaps start by adding a link like this to the README of the main project as a trail, to see if having something like this actually increases the number of outdated icons reported.
> 
> Also the link is incorrect, it should be: `https://github.com/simple-icons/simple-icons/issues/new?labels=icon+outdated&template=icon_update.yml&title=Update%3A+`

Adding this as text is not an option after changing the menu to icons.
Unless it is added outside the menu, somewhere else. 
A different icon could be a megaphone or a flag. 
Adding the link in the Readme is fine. Not sure if one thing supersedes the other.

If it were up to me, the report outdate icon would be an option for each one of the icons, with a little flag, in the footer of the icon card (from left to right, it would be color div, icon for svg download, icon for colored svg download, icon for pdf download, icon to report outdated).

ericcornelissen:
> Adding this as text is not an option after changing the menu to icons.

Agreed that wasn't my suggestion. If it's text it should go somewhere else (as you suggested), probably somewhere else in the `<header>`. Alternatively, if we implement a sidemenu (that can also house the header links on small screens) we could put it in there.


> A different icon could be a megaphone or a flag.

I don't think either of these, in the header, are good alternatives. I think the flag would be a reasonable candidate if we add it to the cards.

> Adding the link in the Readme is fine. Not sure if one thing supersedes the other.

I would consider it unrelated, neither change depends on the other.


> If it were up to me, the report outdate icon would be an option for each one of the icons, with a little flag, in the footer of the icon card (from left to right, it would be color div, icon for svg download, icon for colored svg download, icon for pdf download, icon to report outdated).

You could give it a try, but I think this would add too much HTML to the site right now. But hey, maybe it turns out it barely adds anything compared to the overall site :upside_down_face: 

I do think this would be more realistic when we have pagination or lazy-loading implemented.

jorgeamadosoria:
> I do think this would be more realistic when we have pagination or lazy-loading implemented.

Isn't lazy loading implemented via HTML attribute already?

ericcornelissen:
I meant lazy loading of the HTML content, not just lazy rendering the SVGs.

jorgeamadosoria:
Closing this PR since we discussed adding the report icon to the card rather than the menu.

Issue #50, opened 2021-02-23
Use consistent iconography


ericcornelissen:
### Kind of issue 

- [ ] Feature
- [ ] Bug
- [x] Design
- [ ] Other, namely:

### Description

Currently, the iconography is a bit inconsistent, the icons for ordering were custom made for the old website; the dark/light/system mode icons are taken from [Feather Icons] (the system one is custom-made based on Feather Icons icons); and for #47 we need a "legal" icon which Feather Icon's does not seem to have.

In the table below I aim to create an overview of some icon packs that I'm familiar with and whether or not they have each icon. The most obvious finding, it doesn't look like any icon pack provides "sort by color" or "sort by relevance" :sweat: 

| Icons | [Font Awesome] | [Feather Icons] | [iconsvg] | [Heroicons] | [Phosphor Icons] | [System UIcons] | [Material Design Icons]
| :--- | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| Light mode | [Yes](https://fontawesome.com/icons/sun) | [Yes](https://feathericons.com/?query=sun) | Yes | Yes | Yes | No | Yes |
| Dark mode | [Yes](https://fontawesome.com/icons/moon) | [Yes](https://feathericons.com/?query=moon) | Yes | Yes | Yes | No | Yes |
| Light/Dark mode | No? | No? | No? | No? | No? | No | Yes |
| External link | [Yes](https://fontawesome.com/icons/external-link-alt) | [Yes](https://feathericons.com/?query=external-link) | Yes | Yes | Yes? | Yes? | Yes |
| Copy to clipboard | [Yes](https://fontawesome.com/icons/copy) | [Yes](https://feathericons.com/?query=copy) | Yes | Yes | Yes | Yes | Yes |
| Legal | [Yes](https://fontawesome.com/icons/balance-scale) | No | No | Yes | No | No | Yes |
| Sort alphabetically | [Yes](https://fontawesome.com/icons/sort-alpha-up) | No | No | No? | No? | No? | Yes |
| Sort by color | No? | No | No | No | No | No | Yes? |
| Sort by relevance | No? | No | No | No | No | No |  Yes? |

[Font Awesome]: https://fontawesome.com/
[Feather Icons]: https://feathericons.com/
[Heroicons]: https://heroicons.com/
[iconsvg]: https://iconsvg.xyz/
[Phosphor Icons]: https://phosphoricons.com/
[System UIcons]: https://systemuicons.com/
[...]

Issue #27, opened 2021-01-17
Website header links "GitHub" and "About" are nearly identical


ericcornelissen:
### Kind of issue 

- [x] Improvement
- [ ] Bug
- [ ] Other, namely:

### Description

Currently the _GitHub_ link on the website links to:

```
https://github.com/simple-icons/simple-icons
```

And the _About_ link on the website links to:

```
https://github.com/simple-icons/simple-icons#readme
```

The difference being:

```diff
- https://github.com/simple-icons/simple-icons
+ https://github.com/simple-icons/simple-icons#readme
```

I think it's worth considering removing one of the two. My suggestion would be to remove the _About_ link and possibly change the GitHub link to:

```
https://github.com/simple-icons/simple-icons#readme
```

Discussion and suggestions are welcome 🙂

PeterShaggyNoble:
How about either linking directly to https://github.com/simple-icons/simple-icons/blob/develop/README.md instead or adding an "about" page to the site that pulls in the contents of the README?

ericcornelissen:
For clarity https://github.com/simple-icons/simple-icons#readme links to the same page as https://github.com/simple-icons/simple-icons/blob/develop/README.md (including that, if the default branch changes from `develop` to something else, it still works).

Pulling in the README from [simple-icons/simple-icons](https://github.com/simple-icons/simple-icons) (which is already happinning [through the NPM dependency](https://github.com/simple-icons/simple-icons-website/blob/9a1116bb95d37d98db940e21805385a782e520f9/package.json#L28)) and somehow rendering it as HTML on the website is an interesting idea and would also address #16 :thinking: Not sure how feasible this is but definitely worth considering!

PR #99, opened 2021-08-21
Converted menu links into logos

### LINKED ISSUE #39: [Suggestion] Replace header link text by logos

### LINKED ISSUE #58: Consider design for link to DISCLAIMER.md

jorgeamadosoria:
- Converted menu links to icons:
image

image

Links included now are, in order from left to right:

- Github repository for simple-icons
- npm package
- packagist
- jsDelivr
- Unpkg
- Open Collective
- Legal Disclaimer
- About

Taken mostly from here: https://petershaggynoble.github.io/SI-Sandbox/

Unpkg icon taken (invented?) from https://twitter.com/unpkg
Apparently there is/was a rebranding effort, which is good since Unpkg does not seem to have a logo of their own.

The legal disclaimer icon is a simple scale made wider to allow for light/dark mode switch. However it is visually awful. A better icon would be preferred, specially in dark mode.

The menu is non responsive, as the text menu wasn't either.

@PeterShaggyNoble's sandbox menu is responsive. I think we should copy more things from there, including the npm package command. A responsive menu would be a nice touch.


public/index.pug
```diff
@@ -29,22 +29,40 @@ html(lang="en-US")
...
         li(class="header__list-item")
-          a(href="https://www.jsdelivr.com/package/npm/simple-icons") JSDelivr
+          a(class="header__list-icon" title="jsDelivr (CDN)" href="https://www.jsdelivr.com/package/npm/simple-icons" rel="noopener")
```
ericcornelissen: Wondering if writing "Content Delivery Network" in full here makes more sense, space is not a concern anyway

```suggestion
          a(class="header__list-icon" title="jsDelivr (Content Delivery Network)" href="https://www.jsdelivr.com/package/npm/simple-icons" rel="noopener") 
```

public/index.pug
```diff
@@ -29,22 +29,40 @@ html(lang="en-US")
...
         li(class="header__list-item")
-          a(href="https://unpkg.com/browse/simple-icons/") Unpkg
+          a(class="header__list-icon" title="Unpkg (CDN)" href="https://unpkg.com/browse/simple-icons/" rel="noopener")
```
ericcornelissen: Wondering if writing "Content Delivery Network" in full here makes more sense, space is not a concern anyway

```suggestion
          a(class="header__list-icon" title="Unpkg (Content Delivery Network)" href="https://unpkg.com/browse/simple-icons/" rel="noopener") 
```

public/stylesheet.css
```diff
@@ -41,6 +41,7 @@
   --dm-color-button-text-hover: rgb(0, 0, 0);
   --dm-color-disabled: rgb(170, 170, 170);
   --dm-color-grid-item-divider: rgb(64, 64, 64);
+  --dm-color-menu-icon: rgb(212, 212, 212);
```
ericcornelissen: The buttons are currently missing an effect when hovered, it should probably just be a lighter color (or darker color and make the standard color lighter). Relatedly, I would say the `:focus` styling should be set to the same as the `:hover` styling and have the outline removed.

Also, this color is currently the same as `--dm-color-button-hover` (same for the `--lm` variants) so we should probably re-use those in some way. Do you feel like you can do that? This stylesheet is still a bit of a mess in some regards (including this regard) so I can help you out here if you prefer.

tests/e2e.test.js
```diff
@@ -95,25 +95,22 @@ describe('External links', () => {
...
+    'npm',
+    'packagist',
+    'jsDelivr (CDN)',
+    'Unpkg (CDN)',
+    'Open Collective',
+    'Legal disclaimer',
+    'About Simple Icons',
+  ];
+
+  menuLinks.forEach((title) =>
```
ericcornelissen: Love the improvement 👍 Just one suggestion for clarity:

```suggestion
  const menuLinksTitles = [
    'repository',
    'npm',
    'packagist',
    'jsDelivr (CDN)',
    'Unpkg (CDN)',
    'Open Collective',
    'Legal disclaimer',
    'About Simple Icons',
  ];

  menuLinksTitles.forEach((title) =>
```

public/index.pug
```diff
@@ -29,22 +29,40 @@ html(lang="en-US")
...
         li(class="header__list-item")
-          a(href="https://github.com/simple-icons/simple-icons/blob/develop/DISCLAIMER.md") Disclaimer
+          a(class="header__list-icon" title="Open Collective" href="https://opencollective.com/simple-icons" rel="noopener") 
+            svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
+              path(d="M12 0C5.373 0 0 5.373 0 12s5.373 12 12 12c2.54 0 4.894-.79 6.834-2.135l-3.107-3.109a7.715 7.715 0 1 1 0-13.512l3.107-3.109A11.943 11.943 0 0 0 12 0zm9.865 5.166l-3.109 3.107A7.67 7.67 0 0 1 19.715 12a7.682 7.682 0 0 1-.959 3.727l3.109 3.107A11.943 11.943 0 0 0 24 12c0-2.54-.79-4.894-2.135-6.834z")
+        li(class="header__list-item")
+          a(class="header__list-icon" title="Legal disclaimer" href="https://github.com/simple-icons/simple-icons/blob/master/README.md" rel="noopener") 
+            svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" stroke="#fff")
```
ericcornelissen: I think the `stroke` should be removed here, it [looks a little bit weird in dark mode](https://user-images.githubusercontent.com/15719479/130332670-c5f7b10d-713e-4380-b10d-8ee9ac7d97ed.png)

```suggestion
            svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
```

public/index.pug
```diff
@@ -29,22 +29,40 @@ html(lang="en-US")
...
+          a(class="header__list-icon" title="Unpkg (CDN)" href="https://unpkg.com/browse/simple-icons/" rel="noopener") 
+            svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
+              path(d="M 12.000002,-1.0439e-4 A 12.000002,12.000002 0 0 0 0,11.999898 12.000002,12.000002 0 0 0 12.000002,23.9999 12.000002,12.000002 0 0 0 24.000005,11.999898 12.000002,12.000002 0 0 0 12.000002,-1.0439e-4 Z m 0,2.00178449 A 10.008923,10.008923 0 0 1 22.008925,12.010603 10.008923,10.008923 0 0 1 12.000002,22.019525 10.008923,10.008923 0 0 1 1.9910798,12.010603 10.008923,10.008923 0 0 1 12.000002,2.0016801 Z m 2.140946,3.9286359 -0.03126,4.046388 -0.03329,4.046388 c -0.730063,2.108832 -3.414809,2.108832 -3.992864,0.007 C 10.008601,13.719677 9.9657825,12.23172 9.9657825,9.7482222 V 5.9373383 H 6.8400013 l 0.04314,4.1427307 c 0.053417,4.474577 0.08446,4.742195 0.6615523,5.909011 0.3532561,0.685103 1.2096345,1.445138 1.9696703,1.723461 1.3916151,0.524532 3.6824271,0.524532 5.0740421,0 0.749331,-0.278323 1.552186,-0.975201 1.937556,-1.680642 0.58876,-1.102587 0.654059,-1.627119 0.654059,-6.016059 l 0.0098,-4.0142733 -3.040143,-0.064657 z")
```
ericcornelissen: Looks like this SVG could be optimized with a lower precisions 🤔

public/index.pug
```diff
@@ -29,22 +29,40 @@ html(lang="en-US")
...
         li(class="header__list-item")
-          a(href="https://github.com/simple-icons/simple-icons/blob/develop/DISCLAIMER.md") Disclaimer
+          a(class="header__list-icon" title="Open Collective" href="https://opencollective.com/simple-icons" rel="noopener")
```
ericcornelissen: We _could_ remove it from the footer now, what do you think?

ericcornelissen (changes requested):
Looking good already, nice work. I have some requests for changes but they're mostly minor. I also added to the title that it adds "Open Collective".

The biggest thing I noticed is that it currently doesn't work in Firefox (the icons are not scaled down there), can you look into this?


> The legal disclaimer icon is a simple scale made wider to allow for light/dark mode switch. However it is visually awful. A better icon would be preferred, specially in dark mode.

I agree that a different icon would be preferred (though it being a scale is fine). In the long run it should probably come from the same set as will be used to resolve #50. That said,  I'd be fine merging this PR as is and opening an issue to update the icon (or add it as part of #50).


> I think we should copy more things from there, including the npm package command. A responsive menu would be a nice touch.

I'd recommend you to open (two separate) issues for that - even if you currently plan to resolve them in this Pull Request.

public/stylesheet.css
jorgeamadosoria: I guess I can add them. Agreed on the reuse of button hover and focus

public/index.pug
jorgeamadosoria: agreed. will do

public/index.pug
jorgeamadosoria: true. done.

public/index.pug
jorgeamadosoria: Initially I thought you meant the disclaimer.
I would keep the Open Collective link in both footer and header. It's not going to hurt to have "calls to action" in a couple places. And it's not obtrusive or anything in the footer either, so... I say let it. At least until we need the space for something else more important.

But since you mentioned the footer.... should we add twitter to the menu as well?

ericcornelissen: Just a suggestion (hence the italics _could_), I'm fine either way.

> But since you mentioned the footer.... should we add twitter to the menu as well?

I'm a bit on the fence in that regard: all the other links are to _our_ page on the respective platform (excluding the disclaimer), whereas the twitter button in the footer is to tweet about the project which can at least be confusing and may be interpreted as being misleading (I personally wouldn't like it). What do you think?

public/index.pug
jorgeamadosoria: uff, yes. what an oversight on my part.

public/index.pug
jorgeamadosoria: I agree. It would look out of place in the menu.
Let's keep it down there for the time being

public/stylesheet.css
jorgeamadosoria: added an effect on hover. I also checked it on firefox, it looks ok to me.
The pixel values are totally arbitrary, open to changes if you would like more / less margin or bigger / smaller menu icons

jorgeamadosoria:
Done with the changes. when we have this one merged I will add the report icon button, and after that, I'll create the ticket for responsive menu. 
In that order.

public/index.pug
```diff
@@ -29,22 +29,40 @@ html(lang="en-US")
...
+        li(class="header__list-item")
+          a(class="header__list-icon" title="About Simple Icons" href="https://github.com/simple-icons/simple-icons/blob/master/README.md" rel="noopener") 
+            svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
+              path(d="M12 0C8.688 0 6 2.688 6 6s2.688 6 6 6c1.066 0 2.1171.2862 3.0371.8262 2.858 1.678 3.8167 5.3539 2.1387 8.2129h1.6797a7.4925 7.4925 0 0 0 .6425-3.0293c.003-4.144-3.356-7.5048-7.498-7.5098-2.484 0-4.5-2.016-4.5-4.5S9.516 1.5 12 1.5s4.5 2.016 4.5 4.5H18c0-3.312-2.688-6-6-6zm0 3c-1.597.04-2.8799 1.3259-2.9219 2.9219C9.0351 7.5799 10.343 8.957 12 9c1.597-.04 2.8799-1.3259 2.9219-2.9219C14.9649 4.4201 13.656 3.043 12 3zm0 1.5c.828 0 1.5.672 1.5 1.5v.002c0 .828-.672 1.5-1.5 1.5-.83 0-1.5-.673-1.5-1.502 0-.83.67-1.5 1.5-1.5zM7.5 15v1.5H9v6H4.5V24h15v-1.5H15v-6h1.5V15h-9zm3 1.5h3v6h-3v-6zm-6 1.4707V18a7.418 7.418 0 0 0 .6445 3.039h1.6836C6.3001 20.147 6 19.11 6 18v-.0293H4.5Z")
```
ericcornelissen: What's the motivation for adding this link? I'm not sure what purpose it serves in addition to the GitHub link

public/stylesheet.css
```diff
@@ -264,16 +276,24 @@ a {
...
-.header__list-item:hover {
-  color: var(--color-button-hover);
-  text-decoration: underline;
+
+.header__list-item:hover,
+.header__list-item:focus {
+  cursor: pointer;
+  height: 48px;
+  width: 48px;
+  margin: 6px;
```
ericcornelissen: Interesting choice for a hover/focus effect. I would personally prefer something more similar to existing hover effects (none of which scale the hover target). E.g. change the color of the icon or invert it and add a background. What do you think? If you'd like to stick with what you have now, I think we should ask for input from other maintainers.

public/stylesheet.css
```diff
@@ -264,16 +276,24 @@ a {
...
+}
+
+.header__list-icon {
+  fill: var(--color-menu-icon) !important;
 }
```
ericcornelissen: Am I mistaken or is this rule redundant given [line 284](https://github.com/simple-icons/simple-icons-website/blob/fdf6e6e4a60522671c27832831395cb142de9d70/public/stylesheet.css#L284) :thinking:

ericcornelissen:
Also, is this still a WIP/draft @jorgeamadosoria - if not could you update the title and mark it as ready for review.
Stil a WIP because of the hover effect, which links were going to be added or left in the footer, etc. 
Let's agree on all that first, then we can have it ready for review by anyone else.

public/index.pug
jorgeamadosoria: It's different to point to the repository than to point to the README file.
Besides, the idea is to point to the contributing page, the contributors list, maybe put the legal disclaimer in there as well.

public/stylesheet.css
jorgeamadosoria: I started with the inverse color effect. the problem is that it looks good in the buttons because they have borders. Without borders in the menu logos (which looks ugly) the other choice is to have a round "halo" to serve as background when inverting the color. It's never exactly the same as the buttons hover effect, so this one was faster.
I can go back to inverting the colors, if we are ok living with the circle color to serve as background.
Or any other idea to provide a background for the inverted icon.

public/stylesheet.css
jorgeamadosoria: that depends on whether we are going to go back to inverted colors.

public/index.pug
ericcornelissen: I didn't say it's the same. I'm asking what the point of it is in addition to the existing link. If you want it to point to the readme than we can change the existing link.

If the ultimate goal is to do something else, than I don't see a reason to add this as an intermediate step.

public/stylesheet.css
ericcornelissen: I'm not saying it has to be the same effect as the control buttons, after all there are more interactive elements on the page.

I think the simplest (and best) solution would be to change the color (either make the default gray and hover black or the other way around). But I'm definitely open to suggestions.

public/stylesheet.css
ericcornelissen: So, it is redundant at the moment? It's hard to comment based on a potential change in the future... It's easier to review of the Pull Request is in a consistent state. (Admittedly the PR is still in a draft state but your last comment suggester to me it's ready for review)

jorgeamadosoria: yes, at the moment it is redundant and can be removed. It is in there because I suspected the color inversion was going to be asked to put back in.
That's also why this is a WIP: it's not ready for merging until we iron out all of these details. I wouldn't want accidents to happen again.

public/stylesheet.css
jorgeamadosoria: well, the most visually pretty effect is inverting the colors with the background. But the problem is whether we can live with the circular background.
Or rounded square background, same problem.
Grey could work, but it reduces the contrast with the background color in both light and dark cases, That doesn't help the UX. Better to have high contrast by default (white on black, black on white). Making the hover grey could work, I would have to see how it looks. I can try it.
If you want, I can push a circular background version so you can see it. Maybe I see it being uglier than you, and I'm being too harsh on that variant, idk.

public/index.pug
jorgeamadosoria: the ultimate point of this link, right now, is to point to the readme file rather than the root of the repository.
At a later time, this can be evolved to be a dropdown with more links. Or not. If it stays like this, I'm good.
It's not an intermediate step that has to evolve to something else, it's an intermediate step that can become a final step if we choose to. The menu is stable as it is.
Regardless, I'm not married to it. If you want to remove it, we can. There are more important things to do.

ericcornelissen: I'm just asking what value it adds right now. I'm not saying this should be removed before this PR can be merged. And again I can tell the difference, no need to point that out. If you think it makes sense to have it by all means leave it in. I would like to point your attention to #27 as well.

As for "intermediate" aspect, I personally don't see a reason to add something now because we might want to add something later, while that something we want to add later doesn't _require_ the thing we're adding now.

public/stylesheet.css
ericcornelissen: > the most visually pretty effect is inverting the colors with the background

That is a matter of opinion. (And yes, I also said "the best solution is to change the color", but I prefaced that with "I think".)


> Grey could work, but it reduces the contrast with the background color in both light and dark cases, That doesn't help the UX. Better to have high contrast by default (white on black, black on white). Making the hover grey could work, I would have to see how it looks. I can try it.

As long as the contract is sufficient, I don't see a problem. Don't get me wrong, it's easier to get wrong, but it doesn't have to be.


> If you want, I can push a circular background version so you can see it. Maybe I see it being uglier than you, and I'm being too harsh on that variant, idk.

Only one way to find out, and that's to see what it actually looks like.

public/stylesheet.css
ericcornelissen: I understand now why you leave it as a draft, that makes sense.

However, that does not change the fact that I can only review the changes in the Pull Request as a whole. Again, you cannot expect the reviewer to know that you leave this in because maybe you need it later.

Also, if whether or not this comment is resolved depends on a future change, I don't think it should be marked as resolved yet.

public/index.pug
jorgeamadosoria: The menu now has the background color effect as discussed.

public/index.pug
```diff
@@ -29,22 +29,40 @@ html(lang="en-US")
...
         li(class="header__list-item")
-          a(href="https://github.com/simple-icons/simple-icons/blob/develop/DISCLAIMER.md") Disclaimer
+          a(class="header__list-icon" title="Open Collective" href="https://opencollective.com/simple-icons" rel="noopener") 
+            svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
+              path(d="M12 0C5.373 0 0 5.373 0 12s5.373 12 12 12c2.54 0 4.894-.79 6.834-2.135l-3.107-3.109a7.715 7.715 0 1 1 0-13.512l3.107-3.109A11.943 11.943 0 0 0 12 0zm9.865 5.166l-3.109 3.107A7.67 7.67 0 0 1 19.715 12a7.682 7.682 0 0 1-.959 3.727l3.109 3.107A11.943 11.943 0 0 0 24 12c0-2.54-.79-4.894-2.135-6.834z")
+        li(class="header__list-item")
+          a(class="header__list-icon" title="Legal disclaimer" href="https://github.com/simple-icons/simple-icons/blob/master/README.md" rel="noopener")
```
ericcornelissen: This should link to the disclaimer, not the readme

public/stylesheet.css
```diff
@@ -264,16 +276,27 @@ a {
...
+  margin: 6px;
+  background-color: var(--color-background);
+  fill: var(--color-menu-icon);
 }
-.header__list-item:hover {
-  color: var(--color-button-hover);
-  text-decoration: underline;
+
+.header__list-item:hover,
+.header__list-item:focus {
```
ericcornelissen: This focus doesn't work because the focusable element is the `<a>` (i.e. `.header__list-icon`)

public/stylesheet.css
```diff
@@ -264,16 +276,27 @@ a {
...
+  cursor: pointer;
+  margin: 6px;
+  fill: var(--color-background);
+  background-color: var(--color-menu-icon);
+  border-radius: 10%;
+}
+
+.header__list-icon {
+  fill: inherit;
 }
```
ericcornelissen: This seems redundant, tested it when removed and doesn't seem to affect anything :thinking:

public/stylesheet.css
```diff
@@ -178,6 +188,8 @@ body.light {
   --color-button-text-hover: var(--lm-color-button-text-hover);
   --color-disabled: var(--lm-color-disabled);
   --color-grid-item-divider: var(--lm-color-grid-item-divider);
+  --color-menu-icon: var(--lm-color-menu-icon);
+  --color-menu-icon-hover: var(--lm-color-menu-icon-hover);
```
ericcornelissen: These, and all related variables, are currently unused.

public/stylesheet.css
```diff
@@ -178,6 +188,8 @@ body.light {
   --color-button-text-hover: var(--lm-color-button-text-hover);
   --color-disabled: var(--lm-color-disabled);
   --color-grid-item-divider: var(--lm-color-grid-item-divider);
+  --color-menu-icon: var(--lm-color-menu-icon);
```
ericcornelissen: I think we can and should just reuse the `--color-button` variables for the "menu" item. If you want to have a separate variable - e.g. because we may want a different color in the header at some point - then this variable should for now reference the existing one to ensure they're the same as long as that's the goal.

Also, "menu" is a bit of an odd choice of word as it doesn't appear in the HTML.

ericcornelissen (changes requested):

public/index.pug
ericcornelissen: Okay, but that's unrelated to this comment thread.

public/stylesheet.css
jorgeamadosoria: it is necessary. Otherwise the fill does not go to the svg.
Did you try Chrome? That's what I use.

public/stylesheet.css
jorgeamadosoria: removed color-menu-icon-hover

public/stylesheet.css
jorgeamadosoria: removed color-menu-icon, reused color-button-active

public/stylesheet.css
jorgeamadosoria: I tried firefox, and it works without the fill there. Chrome requires it, however.

public/index.pug
jorgeamadosoria: Link to README has been removed

jorgeamadosoria:
Still something missing here, @ericcornelissen ? I'm only waiting on your review of the changes requested.

public/index.pug
```diff
@@ -29,22 +29,36 @@ html(lang="en-US")
     header(class="header")
       ul(class="header__list")
         li(class="header__list-item")
-          a(href="https://github.com/simple-icons/simple-icons") GitHub
+          a(class="header__list-icon" title="repository" href="https://github.com/simple-icons/simple-icons" rel="noopener")
```
ericcornelissen: Could you remove the `rel="noopener"` here and in the other header links. If there's no `target` specified than it's not needed, see also #83.

public/stylesheet.css
```diff
@@ -264,16 +264,27 @@ a {
...
+  margin: 10px;
+  background-color: var(--color-background);
+  fill: var(--color-button-active);
 }
-.header__list-item:hover {
-  color: var(--color-button-hover);
-  text-decoration: underline;
+
+.header__list-item:hover,
+.header__list-item a:focus {
```
ericcornelissen: This assumes the structure of the document a bit too much I would say, why not just use the `.header__list-icon` selector?

If you think that's not a good solution because styling a class named "icon" in this way is odd, then I'd suggest we introduce a new class for clarity (e.g. `.header__list-link`).

ericcornelissen (changes requested):
The hovering effect seems off to me now without the padding. Why did you change that?

> I'm only waiting on your review of the changes requested.

Sorry, wasn't aware it could be reviewed again as it's still a draft.

public/index.pug
jorgeamadosoria: removed

jorgeamadosoria:
> The hovering effect seems off to me now without the padding. Why did you change that?
> 
> > I'm only waiting on your review of the changes requested.
> 
> Sorry, wasn't aware it could be reviewed again as it's still a draft.

readded padding.

public/index.pug
ericcornelissen: Looks like you removed it everywhere, not only in these header links :confused: 

Also, minor (and unfortunate that Prettier doesn't pick this up), but now there's a space between the last quote and the closing parenthesis of the `.header__list-icon` elements.

jorgeamadosoria: done correcting that.

public/stylesheet.css
```diff
@@ -264,16 +264,27 @@ a {
...
+  width: 55px;
+  background-color: var(--color-background);
+  fill: var(--color-button-active);
 }
-.header__list-item:hover {
-  color: var(--color-button-hover);
-  text-decoration: underline;
+
+.header__list-item:hover,
+.header__list-icon:focus {
```
ericcornelissen: I feel like it would make more sense to use the same selector twice, what do you think @jorgeamadosoria?

```suggestion
.header__list-icon:hover,
.header__list-icon:focus {
```

public/stylesheet.css
```diff
@@ -264,16 +264,27 @@ a {
...
+  background-color: var(--color-button-active);
+  border-radius: 50%;
+}
+
+.header__list-icon {
+  padding: 10px;
+  height: 55px;
+  width: 55px;
+  display: inline-block;
+  fill: inherit;
```
ericcornelissen: Why don't you just set the `fill` value directly here:

```suggestion
  fill: var(--color-button-active);
```

Rather then inheriting it from it's parent [where you specify the color without any effect(?) on line 271](https://github.com/simple-icons/simple-icons-website/blob/50f437a69847dd976ba02619b01f6902680b0cec/public/stylesheet.css#L271).

public/stylesheet.css
```diff
@@ -264,16 +264,27 @@ a {
...
-.header__list-item:hover {
-  color: var(--color-button-hover);
-  text-decoration: underline;
+
+.header__list-item:hover,
+.header__list-icon:focus {
+  cursor: pointer;
+  fill: var(--color-background);
+  background-color: var(--color-button-active);
+  border-radius: 50%;
```
ericcornelissen: This used to be `10%` (e.g. in 2fd961def848b63c93c230566ebec660e029d027), why did you change it to `50%`? This surprised me.

First, I don't think the `50%` looks good with the npm link in particular:

image


More generally speaking, I don't feel like the circle suites the style of the rest of the website. Other elements have slightly rounded corners (e.g. search input, control buttons, cards) [specified in exact pixels](https://github.com/simple-icons/simple-icons-website/blob/cdc99f1be3f68911cbe69b5277daec936e136578/public/stylesheet.css#L24-L25). I would suggest we do the same here for consistency, e.g.:

```suggestion
  border-radius: var(--border-radius-large);
```

What do you think @jorgeamadosoria?

public/stylesheet.css
```diff
@@ -264,16 +264,27 @@ a {
...
+}
+
+.header__list-icon {
+  padding: 10px;
+  height: 55px;
+  width: 55px;
```
ericcornelissen: Is there any particular reason you repeated the exact pixel values here? Why not just do

```suggestion
  height: 100%;
  width: 100%;
```

ericcornelissen (changes requested):

public/stylesheet.css
jorgeamadosoria: it's not possible: header__list-item is a <li> element, it has no focus

public/stylesheet.css
jorgeamadosoria: I don't know, less letters? it's the same effect. 
If you prefer it set directly, we can do it.

public/stylesheet.css
jorgeamadosoria: I don't have strong opinions one way or the other. I put 50% to make the background round. If you prefer it at 6px, that's fine too.

public/stylesheet.css
jorgeamadosoria: no particular reason. I was tweaking the values and it stayed like that.

public/stylesheet.css
ericcornelissen: > I don't know, less letters?

By my count having:

```css
[...]
fill: var(--color-button-active);
[...]
fill: inherit;
[...]
```

Is more letter than 

```css
fill: var(--color-button-active);
```

public/stylesheet.css
ericcornelissen: Just suggested it because now you can't mess up by changing the pixel value only in one place.

public/stylesheet.css
ericcornelissen: If you think `50%` is the way to go, please stick with that and we ship it that way :slightly_smiling_face:

jorgeamadosoria: no, no, it's fine. Like I said, I don't have a horse in this race. Let's go with yours.

jorgeamadosoria: I already committed the change, even.

public/stylesheet.css
jorgeamadosoria: it's fine, I applied the suggestion. It's better not to have duplicates.

jorgeamadosoria:
I think that's all the changes?

public/index.pug
```diff
@@ -29,17 +29,33 @@ html(lang="en-US")
...
         li(class="header__list-item")
-          a(href="https://github.com/simple-icons/simple-icons/blob/develop/DISCLAIMER.md") Disclaimer
+          a(class="header__list-icon" title="Open Collective" href="https://opencollective.com/simple-icons")
+            svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
+              path(d="M12 0C5.373 0 0 5.373 0 12s5.373 12 12 12c2.54 0 4.894-.79 6.834-2.135l-3.107-3.109a7.715 7.715 0 1 1 0-13.512l3.107-3.109A11.943 11.943 0 0 0 12 0zm9.865 5.166l-3.109 3.107A7.67 7.67 0 0 1 19.715 12a7.682 7.682 0 0 1-.959 3.727l3.109 3.107A11.943 11.943 0 0 0 24 12c0-2.54-.79-4.894-2.135-6.834z")
+        li(class="header__list-item")
+          a(class="header__list-icon" title="Legal disclaimer" href="https://github.com/simple-icons/simple-icons/blob/master/README.md")
```
ericcornelissen: Shouldn't this link to the `DISCLAIMER.md` file?

ericcornelissen (changes requested):

public/index.pug
jorgeamadosoria: huh. yes, yes it should. I didn't remove the correct link when removing the duplicate Readme, apparently.

ericcornelissen: Feel free to merge once you fixed it.

ericcornelissen (approved):

PR #102, opened 2021-08-26
Issue 101- lowercased npm in the nav bar

dylvaz:
#101 wanted npm to be lowercased

PR #104, opened 2021-08-26
Resolve problems with unintended package-lock.json changes

### LINKED ISSUE #103: CONTRIBUTING.md - npm install should be replaced with npm ci
As the title says, this will prevent contributors from writing to package-lock.json when installing dependencies and make the contributing experience just a tab bit better 😃.

Some info on npm ci:

    The new npm ci command installs from your lock-file ONLY. If your package.json and your lock-file are out of sync then it will report an error.

    It works by throwing away your node_modules and recreating it from scratch.

    Beyond guaranteeing you that you'll only get what is in your lock-file it's also much faster (2x-10x!) than npm install when you don't start with a node_modules.

dylvaz:
replaced npm install with npm ci in CONTRIBUTING.md #103


ericcornelissen:
We actually recently worked on this problem in the main project (see https://github.com/simple-icons/simple-icons/pull/6179). Unfortunately the solution we used there isn't really applicable here (see also [this comment](https://github.com/simple-icons/simple-icons-website/pull/92#issuecomment-907026877)).

I'm also not sure if recommending the usage of `npm ci` is a good idea. As per [its documentation](https://docs.npmjs.com/cli/v7/commands/npm-ci): "it's meant to be used in automated environments such as test platforms, continuous integration, and deployment". That is, while it solves the problem, it's not how it's "supposed" to be used. Admittedly, the docs are immediately followed by "or any situation where you want to make sure you're doing a clean install of your dependencies". This, arguably, is a reason to put it in the contributing guidelines as you suggested. However, from my point of view it is not because either

1. You understand npm and you know that `npm install` may change the `package-lock.json` file unintentionally so you can use `npm ci` while knowing it's drawbacks.
2. You don't understand and now you think you should always use `npm ci`, even though in many cases you should be using `npm install`.

Ultimately, as I also said in https://github.com/simple-icons/simple-icons/pull/6179, the `package-lock.json` are just a bit of a mess and it's hard to find a suitable solution. The _best_ solution would be for contributors to use the right Node/NPM version, but this is hard to enforce. We can make it easier (add `"engines"` to `package.json`, add an `.nvmrc` file, add an `.npmrc`, provide a containerized dev environment), but I don't really see a perfect solution.

Also, can you revert/fix the changes from #102 that ended up in this Pull Request?

dylvaz:
I've gone ahead and reverted the changes from #102 that ended up in this PR (sorry about that 😅). I understand the reason for not wanting to use ```npm ci```, it could definitely cause contributors to think that ```npm ci``` should be used all the time.

mondeja:
Agreed that is not a good idea, so I'm closing this.

ericcornelissen:
@dylvaz, at least for you, do you think other things are helpful, e.g.:

- A `Dockerfile` for development
- An `.nvmrc` file
- Something other than https://github.com/simple-icons/simple-icons/pull/6179 with an `.npmrc` file
- Specifying the supported npm versions in engines in `package.json`
- Something else entirely

PR #108, opened 2021-09-04
Alleviate package-lock changing problems

ericcornelissen:
Following up on #92, #104, and https://github.com/simple-icons/simple-icons/pull/6179 this is an attempt to alleviate some of the trouble with `package-lock.json` version 1 and 2 (aka npm v6 or lower and npm v7 or higher).

1. First, I specified the desired NPM version (and by extension the desired `package-lock.json` version) in this project's manifest file. (Also fixing https://github.com/simple-icons/simple-icons/pull/6349 for this repo.)
2. Then, I set `engine-strict=true` in a newly added `.npmrc` file for this project. This will make `npm` error if your Node or NPM version does not comply with the engines values of this project or any of our dependencies.
3. Lastly, I added an `.nvmrc` file to allow users of [NVM](https://github.com/nvm-sh/nvm) (which I highly encourage you to pick up you're not using it yet) (also [for Windows](https://github.com/coreybutler/nvm-windows)) to easily switch to the correct NodeJS & npm version for this project.

~~There is one more thing I can think of that we can do, which is to add a `Dockerfile` for development. I'll leave that for a separate Pull Request though, if it is something we want to add.~~ I just realized, [we already have this](https://github.com/simple-icons/simple-icons-website/blob/master/Dockerfile)! And [it's even documented](https://github.com/simple-icons/simple-icons-website/blob/master/CONTRIBUTING.md#using-docker)!

If this works well, we can apply (some of) these changes to other simple-icons projects as well :slightly_smiling_face: 

---

cc @LitoMore & @dylvaz


LitoMore:
I would like to suggest using [Volta](https://volta.sh) instead of NVM.

We could target more tools to specific versions. [Even the npm](https://docs.volta.sh/reference/pin).

ericcornelissen:
@LitoMore, if I'm understanding Volta correctly then the changes in `package.json` are enough to "support" Volta. Correct?

I personally think it's still useful to have an `.nvmrc` file for those that use NVM or asdf (or any other tool that supports this file). But if I'm the only one who thinks so we can omit it from this PR.

jorgeamadosoria:
I don't think any of this is harmful. Can we merge it?

LitoMore:
@ericcornelissen Sorry for the late response. Let's move forward.

package.json
```diff
@@ -61,7 +61,8 @@
     "webpack": "5.51.1",
     "webpack-cli": "4.8.0"
   },
-  "engine": {
-    "node": ">=12.16.0"
+  "engines": {
+    "node": ">=12.16.0",
+    "npm": "<7.0.0"
```
LitoMore: The npm is always binded with Node.js. This requirement will be a little strange. We could use `"node": "^12.16.0",` instead of `>=12.16.0`.

But since npm v8 released. We better not use npm 6, IMO. Developers should keep their npm up to date.

FYI: https://github.com/npm/cli

ericcornelissen: > The npm is always binded with Node.js. This requirement will be a little strange. We could use `"node": "^12.16.0",` instead of `>=12.16.0`.

In that case it should be `"node": "^12.16.0 || ^14.18.0 || ^16.11.0",` as `^12.16.0` only allows using major release v12. I don't see a problem with allowing users to use npm `<7.0.0` with a node version that comes with npm `>7.0.0` since you can downgrade npm if you want (that might be easier for some people than switching to different node version).

> But since npm v8 released. We better not use npm 6, IMO. Developers should keep their npm up to date.

I'd be happy to switch to package-lock v2 in all projects except the main simple-icons project itself given there's a low volume of contributions here - provided other @simple-icons/maintainers agree. If we do change this to `"npm": ">7.0.0",` then I'd suggest we also change `.nvmrc` to v16.

In the main project I'd be more cautious as contributions so far have shown that a lot of contributors are still on npm `<7.0.0` (provided we adopt this strategy in that project of course).

As a last note, I'm not completely in agreement here and until npm v6 is officially deprecated I'd be fine using it. Upgrading, at least to v7 or v8 is as simple as running `npm install` and committing the new `package-lock.json`. As far as a I know the primary advantage of the new lockfile version is efficiency.

LitoMore: Good points. Let's move forward.

LitoMore:
@ericcornelissen They did it https://github.com/npm/cli/releases/tag/v8.1.0.

ericcornelissen:
@LitoMore, are you referring to:

> Add --lockfile-version config option

If so, while ~~I'm more than happy to add it~~ I already added it (see e67c52b251d5db600fa7aea7c72db192303c2a64), I don't think it actually alleviates any problems. Just like there are people using npm pre-7.0.0 and not upgrading, there will also be people using npm pre-8.1.0 and not upgrading.

If you were referring to something else, please clarify :slightly_smiling_face:

jorgeamadosoria:
This looks like it's ready for what it wants to achieve. What are we waiting to merge it?

ericcornelissen:
If you feel this is good you can merge it @jorgeamadosoria, just don't want to merge it myself

jorgeamadosoria:
it's perfectly fine for what you want to achieve.
Only one concern (excuse my ignorance): the config says < 7.0.0, so that means only version 6 of npm?
Why not <= 7.0.0? or does < 7.0.0 also includes npm7?

ericcornelissen:
`<7.0.0` does **not** include v7.0.0 (allowing only v7.0.0 but not any other v7 release would be an odd choice). As per the discussions, to ensure package-lock version 1 is used we disallow any version that only supports package-lock version 2 (following [this comment](https://github.com/simple-icons/simple-icons-website/pull/108#issuecomment-943942328), starting v8.1.0 you can choose the package-lock version).

If you want to go with `>=7.0.0` and use package-lock version 2, @jorgeamadosoria, feel free to adopt this PR or create a new PR. The only advice I'd give is to keep it consistent within the simple-icons organisation.

jorgeamadosoria:
> `<7.0.0` does **not** include v7.0.0 (allowing only v7.0.0 but not any other v7 release would be an odd choice). As per the discussions, to ensure package-lock version 1 is used we disallow any version that only supports package-lock version 2 (following [this comment](https://github.com/simple-icons/simple-icons-website/pull/108#issuecomment-943942328), starting v8.1.0 you can choose the package-lock version).
> 
> If you want to go with `>=7.0.0` and use package-lock version 2, @jorgeamadosoria, feel free to adopt this PR or create a new PR. The only advice I'd give is to keep it consistent within the simple-icons organisation.

I don't know enough about this config to make any changes, just wanted to make sure that npm 7 was meant to be excluded on purpose.
If that's the case, then I have no issues with this PR. We should merge it and move to other things.

PR #109, opened 2021-09-04
Set save-exact in .npmrc to enforce --save-exact

ericcornelissen:
I just learned about the [`save-exact` option](https://docs.npmjs.com/cli/v7/using-npm/config#save-exact) while working on #108 and figured we can add it to the `.npmrc` file to enforce that dependencies are always installed at an exact version - which is how we usually want to install dependencies in the simple-icons organization.

If this gets merged here, we probably want to apply this to all the repositories.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+save-exact=true


jorgeamadosoria:
This sounds good to me. Can we merge it?

ericcornelissen:
I'll merge it given your approval @jorgeamadosoria  :slightly_smiling_face:

Issue #86, opened 2021-07-04
Tabbing order confusion


ericcornelissen:
### Kind of issue 

- [ ] Feature
- [ ] Bug
- [ ] Design
- [x] Other, namely: a11y

### Description

Currently, tabbing through the website really only works if you're using the default ordering (alphabetical). If either colored ordering or relevance ordering (when searching) is used, tabbing through the grid items will result in jumping all over the page.

This is because the visual re-ordering of the grid items is achieved using the flexbox `order` property and dynamically assigned CSS variables. The idea behind this approach, when it was implemented, was to leverage the GPU to do the re-ordering.

Unfortunately, this is a minor (?) problem for accessibility and if performance isn't too bad when shuffling the elements around to preserve a sane tab ordering we should change the implementation.

developStorm:
Can we switch between different [`tabindex`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) when changing ordering?

ericcornelissen:
Thanks for the suggestion @developStorm, I excluded it originally because I always heard you _shouldn't_ set it to a value other than `-1` and `0` but now I agree that this seems like a decent use case. That said, I still think the first thing we should look into is just moving the DOM elements around.

jorgeamadosoria:
> Thanks for the suggestion @developStorm, I excluded it originally because I always heard you _shouldn't_ set it to a value other than `-1` and `0` but now I agree that this seems like a decent use case. That said, I still think the first thing we should look into is just moving the DOM elements around.

moving DOM elements around will make #35 (EDIT: Meant #34) very hard.

ericcornelissen:
Did you mean #34 @jorgeamadosoria?

jorgeamadosoria:
> Did you mean #34 @jorgeamadosoria?

yes, #34

ericcornelissen:
I'd say this issue definitely take precedence over #34, if you're concerned they will affect each other.

However, I'm not sure if I agree as I would tackle #34 (and #35 for that matter) at built time, but that can be discussed in the respective threads.

PR #110, opened 2021-09-06
Custom color sorting implementation to remove black icons from ATF icon list

### LINKED ISSUE #66: Some blacks are not correctly ordered when ordering by color
### Kind of issue 

- [ ] Feature
- [x] Bug
- [ ] Design
- [ ] Other, namely:

### Description

When ordering by colour the ordering is a bit off as some black colors are displayed at the top, before the reds, while most blacks are displayed all the way at the bottom, following the whites.

Ideally, we find some sort of solution for this ordering problem. Unfortunately, in general there isn't necessarily a good solution for sorting colors. There will always be some cutoff point and it will probably never be perfect.

However, with some effort we may be able to improve the current situation. Two (possible) goals:

- Group all "blacks" and "whites" at the end.
- Have at least the ordering above the fold look "as expected".

Anyone looking to work on this issue, share any suggestions, research, or previews you can :slightly_smiling_face: 

### Example

Below is an example of what ordering by colour looks like - using [v4.8.0 of simple-icons](https://github.com/simple-icons/simple-icons/tree/4.8.0). As you can see _"GitHub"_ and _"DS Autombiles"_, which have dark grey colours, appear before _"The Spriters Resource"_ which is red, i.e.

`#181717` `#1D1717` `#BE3939`

image

### LINKED ISSUE #82: Add integration test case for "Order by color"
### Kind of issue 

- [ ] Feature
- [ ] Bug
- [ ] Design
- [x] Other, namely: Testing

### Description

Relates to #66.

We should add an integration test verifying that the "Order by color" button changes the ordering of the icons. There already exists [a prototype test case](https://github.com/simple-icons/simple-icons-website/blob/ea9452ce56c77f80dfaff4ecf99de22ccd192859/tests/integration.test.js#L297-L307) for this, but it's skipped as it currently doesn't work.

jorgeamadosoria:
As discussed in #66,  the `color-sorter` implementation puts some black shades at the beginning of the color list. After some testing, I've pieced together a color implementation that approximates the requirements of #66.

We would need to phase out the `color-sorter` dependency and perhaps tweak some parameters for the algorithm, namely:
- **Grey range**: set to +/-10. A grey color is defined as an RGB value of the form R=G=B.  The grey range variable defines the variance allowed for R,G,B values that will still be considered "equal" for the purposes of grey detection.
- **Black cutoff**: set to 15. Black cutoff defines the luminosity (in HSL format) for dark colors to be considered "black". The lower it goes, the darker the color.
- **White cutoff**: set to 80. White cutoff defines the luminosity (in HSL format) for light colors to be considered "white". The higher it goes, the lighter the color.

The code may accept some optimization, if someone's willing to go over it.


public/scripts/utils.js
```diff
@@ -44,4 +44,99 @@ module.exports = {
...
+        b >= g - range &&
+        b <= g + range &&
+        r >= b - range &&
+        r <= b + range
+      );
+    };
+
+    const rgbToHsl = (rgb) => {
+      var { r, g, b } = { ...rgb };
+      // Make r, g, and b fractions of 1
```
ericcornelissen: Isn't this called normalizing?

```suggestion
      // Normalize r, g, and b
```

If you choose this wording because _normalize_ is a "difficult" word, I don't think it helps anyone to describe it this way.

public/scripts/utils.js
```diff
@@ -44,4 +44,99 @@ module.exports = {
       .normalize('NFD')
       .replace(/[\u0300-\u036f]/g, '');
   },
+
+  sortByColors: function (colors) {
```
ericcornelissen: This should not be part of this file, this file contains utility functions for the website when displayed in the browser. This function will not be used when the website is live.

It should probably go in a file in [the `scripts` directory](https://github.com/simple-icons/simple-icons-website/tree/00139387a7133ce1eadd86eb70a0e28e8f35dac4/scripts) ([following the pattern of the main repo](https://github.com/simple-icons/simple-icons/blob/91a945820677dfff652920cac065669b350e194b/scripts/utils.js)).

public/scripts/utils.js
```diff
@@ -44,4 +44,99 @@ module.exports = {
...
+      // Red is max
+      else if (cmax == r) h = ((g - b) / delta) % 6;
+      // Green is max
+      else if (cmax == g) h = (b - r) / delta + 2;
+      // Blue is max
+      else h = (r - g) / delta + 4;
+
+      h = Math.round(h * 60);
+
+      // Make negative hues positive behind 360°
```
ericcornelissen: I don't get the "behind 360" part of this comment. Negative hues will never be greater than 360 after the following line. Maybe just say:

```suggestion
      // Ensure all hues are positive
```

public/scripts/utils.js
```diff
@@ -44,4 +44,99 @@ module.exports = {
...
+      if (h < 0) h += 360;
+      // Calculate lightness
```
ericcornelissen: Not sure, but should this be "luminosity"? (also on line 134)

```suggestion
      // Calculate luminosity
```

public/scripts/utils.js
```diff
@@ -44,4 +44,99 @@ module.exports = {
...
+      l = (cmax + cmin) / 2;
+
+      // Calculate saturation
+      s = delta == 0 ? 0 : delta / (1 - Math.abs(2 * l - 1));
+
+      // Multiply l and s by 100
```
ericcornelissen: I'd say either remove this comment or explain why you multiply them by 100

tests/e2e.test.js
```diff
@@ -260,7 +261,7 @@ describe('Search', () => {
 describe('Ordering', () => {
   const icons = Object.values(simpleIcons);
   const titles = icons.map((icon) => icon.title);
-  const hexes = icons.map((icon) => icon.hex).sort(sortColors);
+  const hexes = sortByColors(icons.map((icon) => icon.hex));
```
ericcornelissen, line 262: It's fine either way but any particular reason you didn't implement `sortByColors` so that it can be used as `arr.sort(sortByColors)`?

ericcornelissen (changes requested):
Looking good already. Just a few comment and questions to help me (and hopefully others) better understand what's going on.

> and perhaps tweak some parameters for the algorithm

Probably, I'll play around with the parameters later and share my findings :slightly_smiling_face: 

> The code may accept some optimization, if someone's willing to go over it.

Doesn't seem necessary to me in general (except for [this comment](https://github.com/simple-icons/simple-icons-website/pull/110#discussion_r702661049)), it's just a build step after all - as long as it's not _too_ slow it should be fine.

public/scripts/utils.js
jorgeamadosoria: expanded the comment

tests/e2e.test.js
jorgeamadosoria: I need full control of the array to distribute them in buckets. The sort function only takes two elements at a time.

It may be possible to implement it as a stateless sorting function, but that was not the way it started. If it is absolutely necessary I can try it, but it will be significantly harder than what is already done.

ericcornelissen: Makes sense, in that case just leave it as is.

public/scripts/utils.js
jorgeamadosoria: moved to its own file in the scripts folder

scripts/color-sorting.js
```diff
@@ -0,0 +1,101 @@
...
+      l = +(l * 100);
+      return { h, s, l };
+    };
+
+    const GREY_RANGE = 10;
+    const BLACK_CUTOFF = 15;
+    const WHITE_CUTOFF = 80;
+    var colored = [];
+    var bgw = [];
+    var mappedColors = colors.map((c) => {
```
ericcornelissen: This variable is unused

scripts/color-sorting.js
```diff
@@ -0,0 +1,101 @@
...
+      s = delta == 0 ? 0 : delta / (1 - Math.abs(2 * l - 1));
+
+      // Multiply l and s by 100 to get the value in percent, rather than [0,1]
+      s = +(s * 100);
+      l = +(l * 100);
+      return { h, s, l };
+    };
+
+    const GREY_RANGE = 10;
+    const BLACK_CUTOFF = 15;
```
ericcornelissen: Of course this is very subjective, but I think we should go with a lower value. [The old website used `10` as a cutoff point](https://github.com/simple-icons/simple-icons/blob/2ce6c91de72f09049989b166c292527d33cd5e05/index.html#L148) and, at least to my eye, that looks like a better cut off point than `15`:

```suggestion
    const BLACK_CUTOFF = 10;
```

scripts/color-sorting.js
```diff
@@ -0,0 +1,101 @@
...
+    const WHITE_CUTOFF = 80;
```
ericcornelissen: The old website didn't have a cutoff point for whites, and to be honest I don't feel like we need it - at least with the current set of colors. I definitely think `80` is too low and it's putting some colors with the whites that I wouldn't expect to be there. I would be fine with setting it to either 90 (for symmetry) or anything above that, which currently doesn't put any color in this bucket.

```suggestion
    const WHITE_CUTOFF = 90;
```

scripts/color-sorting.js
```diff
@@ -0,0 +1,101 @@
+/**
+ * @fileoverview
+ * Color sorting algorithm.
```
ericcornelissen, line 3: If it only provides (one) color sorting algorithm, I would not expect the export to be an object. I suggest one of three things:

- Export only the `sortByColors` function
- Make this file more general (e.g. just `utils.js`)
- Update the `@fileoverview` to be plural (which is a bit weird as I don't think we'll be having more than one)

scripts/color-sorting.js
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+module.exports = {
+  sortByColors: function (colors) {
+    const isGray = (rgb, range) => {
+      var { r, g, b } = { ...rgb };
```
ericcornelissen: Could you replace all uses of `var` by `let`? ([see this article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#description))

ericcornelissen (changes requested):
Just a couple of small comments.

scripts/color-sorting.js
jorgeamadosoria: This is subjective, as you said. We go with whatever you choose.

scripts/color-sorting.js
jorgeamadosoria: we need it to remove the whites from general ordering of color icons.
We could put the threshold to 90, but be warned that while that will remove the light magenta and light pink, it would also leave a lot of white colors sprinkle throughout the general color array.

scripts/color-sorting.js
jorgeamadosoria: exported only the sortByColors function

scripts/color-sorting.js
jorgeamadosoria: sure.

ericcornelissen (approved):

jorgeamadosoria:
@ericcornelissen thanks for approving. I won't merge this PR until I get the test case to work too, that way we can close two issues and have a full coding unit

jorgeamadosoria:
> @ericcornelissen thanks for approving. I won't merge this PR until I get the test case to work too, that way we can close two issues and have a full coding unit

the test case(s) have been added.

@ericcornelissen care to take a look?

tests/e2e.test.js
```diff
@@ -295,17 +296,34 @@ describe('Ordering', () => {
...
+      const $color = await $button.evaluate((el) =>
+        JSON.stringify(el.textContent),
+      );
+
+      await expect($color).toMatch(`#${hexes[$idx]}`);
     }
   });
+
+  // issue: https://github.com/simple-icons/simple-icons-website/issues/66
+  it('compares the first two colors to make sure they are not black', async () => {
```
ericcornelissen: Minor, this description doesn't follow the "style" of the rest of the tests. It describes what the tests does rather than what the test tests. A more "in style" description could be:

```suggestion
  it('doesn't have a black-like color at the start', async () => {
```

That said, if you think the current description is better, no need to change it.

tests/e2e.test.js
```diff
@@ -295,17 +296,34 @@ describe('Ordering', () => {
...
+    // should not match black colors
+    await expect(hexes[0]).not.toMatch('181717');
+    await expect(hexes[1]).not.toMatch('1D1717');
+
+    // should match red colors
+    await expect(hexes[0]).toMatch('BE3939');
+    await expect(hexes[1]).toMatch('CB3837');
```
ericcornelissen: I think these tests are way to specific and it's better to just remove it.

The first two asserts only verify these exact two colors don't appear there, not accounting for other colors. Even if these are the only two colors that can appear here, it assumes way too much about implementation in my opinion.

The last to asserts break as soon as there's a brand that has a color that should appear before "The Spriters Resource". Again, even if these must always be the first two colors (which is actually impossible, as there may be two brands that have `#BE3939`), it assumes way too much about implementation in my opinion.

tests/e2e.test.js
```diff
@@ -295,17 +296,34 @@ describe('Ordering', () => {
     }
   });
 
-  it.skip('orders grid items by color', async () => {
-    // Items are ordered using the CSS property `order`...
-    await expect(page).toClick('#order-color');
+  it('orders grid items by color', async () => {
```
ericcornelissen: The tests currently assume the tests are executed in order, every tests that tests something related to color ordering should click the `#order-color` button first.

tests/e2e.test.js
```diff
@@ -295,17 +296,34 @@ describe('Ordering', () => {
...
 
-    const $gridItems = await page.$$('.copy-color');
-    for (let i = 0; i < $gridItems.length; i++) {
-      const $gridItem = $gridItems[i];
-      const hex = hexes[i];
-      await expect($gridItem).toMatch(`#${hex}`);
+      const $text = await $items[i].evaluate((el) =>
+        JSON.stringify(el.getAttribute('style')),
+      );
+      const $idx = parseInt($text.substring(15, $text.length - 1));
```
ericcornelissen: You can use the [`getAttribute` helper](https://github.com/simple-icons/simple-icons-website/blob/00139387a7133ce1eadd86eb70a0e28e8f35dac4/tests/helpers.js#L14-L20) here.

```suggestion
      const $text = await getAttribute($items[i], 'style');
      const $idx = parseInt($text.substring(14, $text.length));
```

Also, I don't understand why you're `JSON.stringify`ing the result of `getAttribute()` here.

tests/e2e.test.js
```diff
@@ -295,17 +296,34 @@ describe('Ordering', () => {
...
+
+      const $color = await $button.evaluate((el) =>
+        JSON.stringify(el.textContent),
+      );
```
ericcornelissen: You may convert this into a helper as well. E.g. `getTextContent($el)`.

tests/e2e.test.js
```diff
@@ -295,17 +296,34 @@ describe('Ordering', () => {
...
+
+      await expect($color).toMatch(`#${hexes[$idx]}`);
```
ericcornelissen: This test currently verifies whether the color of the card matches the color at the index specified by the CSS variable `--order-color` for that card in the `hexes` array. I don't think it verifies whether the ordering is displayed correctly as it tests wither `--order-color` is set correctly. Yes these are related, yes these may be better than no test (I personally disagree, as it gives a false sense that everything is okay), but it doesn't test whether the human viewing the screen will see the grid items sorted by color.

@jorgeamadosoria, the test wasn't skipped because I couldn't come up with some test. It's skipped because I couldn't come up with a test that verifies the color sorting works **end to end**. This is likely impossible (or at least very, very hard) due to the use of the `order var(--order-color)` approach and will likely need a fix related to #86 to be testable.

If you think this test is useful, I'm happy to approve it. I want to stress though that it is important to remember that it is not testing, **end to end**, whether or not color sorting is working.

ericcornelissen (changes requested):

tests/e2e.test.js
jorgeamadosoria: no, you are right. Changing it.

tests/e2e.test.js
jorgeamadosoria: yes, that is correct. That's the way the test should be: very specific.
Perhaps it's too specific for end to end testing, and this would be better in an unit test suite. Do we have one of those?
The idea of the test case is to be specific, and to break easily. Perhaps the last two is too much detail, I would agree on that, but the first two are certainly important to make sure the issue in the comment is addressed.And they appear again, it's either because they are the lightest colors in the collection (highly unlikely) or there is an error, in which case the test should break.

In short: 
- The case is specific by design.
- The last two matches may be too specific, I'll be happy to remove them. The first two however I think are good where they are.
- This may be more suited for unit testing than e2e. Please advice if we have a more fitting place to put this testcase.

tests/e2e.test.js
jorgeamadosoria: no, the test does not assume any order. The test bypasses the need to click on the order-color: the click on the order color does nothing testable, since the css order property does not change the DOM elements order.
Notice that the test takes the grid cards and reads the --style-color attribute, and makes sure that the value assigned during build is the same as the index of the corresponding hex in the sortByColor call in the unit test.

There is no need for any previous test to be executed.

tests/e2e.test.js
ericcornelissen: I agree that sometimes specific tests are important, the reason I disagree in this case is that we're testing something very specific about something that is not in our control. This test can be broken by a non-breaking change in one of our dependencies!

tests/e2e.test.js
ericcornelissen: In that case, as per one of my other comments, it's not actually testing if the `#order-color` button does it job, which is what this (or at the very least one of these tests) should do.

ericcornelissen: See also #82:

> We should add an integration test verifying that the "Order by color" button changes the ordering of the icons.

tests/e2e.test.js
jorgeamadosoria: the test goes as far as it can. There is no way to test the way the page is going to look.
However, the css property uses order just fine, so checking that order is properly generated is enough. 
There is no other way of testing this. The DOM does not change. 
If you want to move the DOM around as discussed in #86, then this test can be amended, but as the coding stands right now, this is as good as it gets.

tests/e2e.test.js
jorgeamadosoria: there is no way to test that. It's impossible since CSS does not change the order of the DOM elements.
I even tried downloading the page HTML after ordering, and it gets downloaded in alpha sort order.

We can:
- either close this as won't add.
- Leave the test as is, to make sure --order-color is being generated as expected
- Implement #86 by moving around DOM elements, and try again.

ericcornelissen: I agree that it is likely impossible to test this with our current setup given that the DOM order doesn't change.

If we want to test that `--order-color` is being generated as expected, then I don't think it should be an end-to-end test. Again, you're not testing something the users sees. For example, I can easily break the website without breaking this test by removing [this line of CSS](https://github.com/simple-icons/simple-icons-website/blob/00139387a7133ce1eadd86eb70a0e28e8f35dac4/public/stylesheet.css#L704). That's what I mean when [I say](https://github.com/simple-icons/simple-icons-website/pull/110#discussion_r706635186) that "it gives a false sense that everything is okay".

Implementing the DOM-reordering suggestion for #86 would indeed make this actually end-to-end testable. Having this tested would be the preferred option. I'd personally be against closing it as "won't add" because we can't think of a solution now and we _want_  to reach 0 open issues, but if you think that makes sense, again, feel free to close those issues.

---

Also, from [this comment](https://github.com/simple-icons/simple-icons-website/pull/110#discussion_r706661196)

> but as the coding stands right now, this is as good as it gets.

I want to stress this once more, I think not adding this test is better than adding this test because it gives a false sense of correctness. That may just be my opinion, and if you disagree I'm happy to merge this in as is.

tests/e2e.test.js
ericcornelissen: This discussion is nearly identical to [this one](https://github.com/simple-icons/simple-icons-website/pull/110#discussion_r706673371), let's keep our focus there.

tests/e2e.test.js
jorgeamadosoria: I'm going to remove the test, then.
However, #86 implementation is not going to be done anytime soon. It's not extremely hard, but the performance hit is going to be painful.

jorgeamadosoria: Test is removed. shall we merge the algorithm?

ericcornelissen: I'm not saying this shall not be merged unless the test is disabled. I'm only sharing my thoughts, [I'm not commanding you to do anything](https://github.com/simple-icons/simple-icons-website/pull/110/commits/bbe6e645138b026447eb2fac7db36fae2a182071). I'm simply sharing my believe that in this case no test may be better than some test.

If you think my reviews are to fine grained, don't ask me for a review. If you think my reviews are slowing down development to much, don't ask for my reviews (as I'm of the opinion that for a project like this, quality goes over quantity)). If you think this test is slowing down the bugfix getting merged, then I refer you to [this comment](https://github.com/simple-icons/simple-icons-website/issues/82#issuecomment-914564965).

tests/e2e.test.js
jorgeamadosoria: how? the only way to break this is to have those two colors close to black appearing at the beginning of the list (notice that I removed the last two assertions).

for that to happen it require changes to the color srting algo or removing thousands of icons.

also, this is independent of all dependencies, AFAIK. Which dependency affect this test?

ericcornelissen: What if GitHub's color changes from `#181717` to `#191717`? Then `await expect(hexes[0]).not.toMatch('181717');` is pointless.

> notice that I removed the last two assertions

That is already better. My last comment was indeed mostly focused on the last two assertions.

jorgeamadosoria: yes, if it changes the test will be pointless. However the idea of the unit test is to avoid regressions. If the color changes and the test case is found to be pointless, **then** it can be eliminated. That's the nature of unit tests, to evolve as conditions change.

ericcornelissen: > If the color changes and the test case is found to be pointless, then it can be eliminated.

And how will you know, in the context of this repository, when it changes? If the `hexes` were part of the source code I would have no objections.

If you want to actually unit test this, then you should create a unit test suite for the script you added (`scripts/color-sorting.js `) and test it with a fixed list of hexes. As long as your test depends on the simple-icons dependency it is not a unit test, it is at least an integration test.

> That's the nature of unit tests, to evolve as conditions change.

What this means is that if the implementation of the sorting algorithm changes because we want a different order, than we should update the test as well. Not, that we should update the test because the input "unexpectedly" changes.

jorgeamadosoria: I've moved the testing case to ordering.test.js and added a mock color array to make the input fixed.
Check it, plz.

tests/ordering.test.js
```diff
@@ -153,4 +153,39 @@ describe('Ordering', () => {
       storedValue,
     );
   });
+
+  // issue: https://github.com/simple-icons/simple-icons-website/issues/66
+  it('does not have a black-like color at the start', async () => {
```
ericcornelissen: This test doesn't have to be async

tests/ordering.test.js
```diff
@@ -153,4 +153,39 @@ describe('Ordering', () => {
...
+      'A100FF',
+      '26689A',
+      'A9225C',
+      '83B81A',
+      '0085CA',
+      '0B2C4A',
+    ];
+
+    const hexes = sortByColors(colors);
+    console.log(hexes);
```
ericcornelissen: Probably shouldn't be logging in a test

tests/ordering.test.js
```diff
@@ -153,4 +153,39 @@ describe('Ordering', () => {
...
+    // should not match black colors
+    await expect(hexes[0]).not.toMatch('181717');
+    await expect(hexes[1]).not.toMatch('1D1717');
```
ericcornelissen: Why not test if output of the sorting algorithm against an expected list of hexes, rather than just checking if the first and second are not some particular elements in the list?

At that point, the comment referring to #66 also becomes obsolete as having a test like this just makes sense I would say. But you may keep it around if you think that reference is useful.

tests/ordering.test.js
```diff
@@ -153,4 +153,39 @@ describe('Ordering', () => {
...
+
+    await expect(hexes[hexes.length - 1]).toMatch('000000');
```
ericcornelissen: If you disagree with the comment above, it's more expressive to use `toEqual` as you want to know if it equals, not if it matches.

```suggestion
    await expect(hexes[hexes.length - 1]).toEqual('000000');
```

tests/ordering.test.js
jorgeamadosoria: this was an oversight. Removing.

tests/ordering.test.js
jorgeamadosoria: I will keep the reference around, and also change to check the full output.

tests/ordering.test.js
jorgeamadosoria: yes, toEqual is the way to go for the full array anyways. Since we are comparing the whole output, this assertion goes away as superfluous

jorgeamadosoria: Done, with these suggestions. Very useful.

tests/ordering.test.js
```diff
@@ -153,4 +153,59 @@ describe('Ordering', () => {
...
+      'EF2D5E',
+      '41454A',
+      '1D1717',
+      '191A1B',
+      '181717',
+      '000000',
+    ];
+
+    const hexes = sortByColors(colors);
+    // should not match black colors
```
ericcornelissen: This comment seems a bit outdated :sweat_smile:

ericcornelissen (approved):
Just one minor thing, then this is good to merge :+1:

tests/ordering.test.js
jorgeamadosoria: ups. yes it is, my bad.

PR #111, opened 2021-09-11
Adding link to website repository in the footer

### LINKED ISSUE #49: Add link pointing to simple-icons-website repository

jorgeamadosoria:
Added a link to the website repository on github:

image

image

### DIFF
--- a/public/index.pug
+++ b/public/index.pug
@@ -29,7 +29,7 @@ html(lang="en-US")
     header(class="header")
       ul(class="header__list")
         li(class="header__list-item")
-          a(class="header__list-icon" title="repository" href="https://github.com/simple-icons/simple-icons")
+          a(class="header__list-icon" title="main repository" href="https://github.com/simple-icons/simple-icons")
             svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
               path(d="M12 .297c-6.63 0-12 5.373-12 12 0 5.303 3.438 9.8 8.205 11.385.6.113.82-.258.82-.577 0-.285-.01-1.04-.015-2.04-3.338.724-4.042-1.61-4.042-1.61C4.422 18.07 3.633 17.7 3.633 17.7c-1.087-.744.084-.729.084-.729 1.205.084 1.838 1.236 1.838 1.236 1.07 1.835 2.809 1.305 3.495.998.108-.776.417-1.305.76-1.605-2.665-.3-5.466-1.332-5.466-5.93 0-1.31.465-2.38 1.235-3.22-.135-.303-.54-1.523.105-3.176 0 0 1.005-.322 3.3 1.23.96-.267 1.98-.399 3-.405 1.02.006 2.04.138 3 .405 2.28-1.552 3.285-1.23 3.285-1.23.645 1.653.24 2.873.12 3.176.765.84 1.23 1.91 1.23 3.22 0 4.61-2.805 5.625-5.475 5.92.42.36.81 1.096.81 2.22 0 1.606-.015 2.896-.015 3.286 0 .315.21.69.825.57C20.565 22.092 24 17.592 24 12.297c0-6.627-5.373-12-12-12")
         li(class="header__list-item")
@@ -176,5 +176,5 @@ html(lang="en-US")
         span Share this on Twitter!
       div(class="footer-repo")
           p.
-            #[a(title="github repository" href="https://github.com/simple-icons/simple-icons-website") Made with &#10084; on GitHub]
+            #[a(title="website repository" href="https://github.com/simple-icons/simple-icons-website") Made with &#10084; on GitHub]
     input(id="copy-input" class="hidden" aria-hidden="true")
--- a/public/index.pug
+++ b/public/index.pug
@@ -177,5 +177,5 @@ html(lang="en-US")
         span Share this on Twitter!
       div(class="footer-repo")
           p. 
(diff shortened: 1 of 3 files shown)


ericcornelissen (commented):
Should we add an end-to-end test for this as well?

jorgeamadosoria:
> Should we add an end-to-end test for this as well?

not sure. If you think it warrants it, sure.

IMO, it's just a link, there is little to test automatically. But it can be done if you want, I'm not against it per se.

jorgeamadosoria:
I added a test case in the same test as the menu, since that test is for "external links".

public/index.pug
```diff
@@ -175,5 +175,7 @@ html(lang="en-US")
           title Twitter logo
           path(d=twitterIcon.path)
         span Share this on Twitter!
-
+      div(class="footer-repo")
+          p. 
+            #[a(title="github repository" href="https://github.com/simple-icons/simple-icons-website") Made with ❤️ on GitHub]
```
ericcornelissen: Is `❤️` a [UTF-8](https://github.com/simple-icons/simple-icons-website/blob/6e919f786dd4413327113b2099a88524c7c99e96/public/index.pug#L4) character?

public/stylesheet.css
```diff
@@ -748,6 +748,7 @@ body.order-by-relevance .grid-item {
   .footer .footer-share {
     flex-grow: 1;
     justify-content: center;
+    margin-bottom: 0.3rem;
```
ericcornelissen: I feel like these are still a bit close together on smaller screens: 

image

Maybe we can increase the value a bit here, e.g.

```suggestion
    margin-bottom: 0.8rem;
```

public/index.pug
```diff
@@ -175,5 +175,7 @@ html(lang="en-US")
           title Twitter logo
           path(d=twitterIcon.path)
         span Share this on Twitter!
-
+      div(class="footer-repo")
+          p.
```
ericcornelissen: Minor but could you remove the trailing space here before someone else's editor (e.g. mine) removes it? (I really thought Prettier would take care of this :unamused:)

```suggestion
          p.
```

jorgeamadosoria: yes, I don't know what happened there.

public/index.pug
jorgeamadosoria: It's a HTML entity now.

tests/e2e.test.js
```diff
@@ -110,6 +110,12 @@ describe('External links', () => {
       await expect(page).toClick('a', { title });
     }),
   );
+
+  const footerRepositoryTitle = 'github repository';
+
+  it(`is possible to click the link for Github repository`, async () => {
+    await expect(page).toClick('a', { footerRepositoryTitle });
```
ericcornelissen: This test doesn't seem to work, changing `footerRepositoryTitle` doesn't affect it on my end. Currently, it seems to just tests if there is _any_ link that can be clicked. (and, actually, the same goes for line 110).

You can find the documentation for [`toClick` here](https://github.com/smooth-code/jest-puppeteer/blob/master/packages/expect-puppeteer/README.md#expectinstancetoclickselector-options). I assumed, `{ title }` would look at the `title` attribute, but it seems that's not a thing that's supported.

While we're talking about this, it's not quite clear to me why you define `footerRepositoryTitle` outside the scope of the test and why you use backticks for the test name.

In this case you can fix the test using, e.g.:

```suggestion
    await expect(page).toClick('a', { text: /Made with . on GitHub/ });
```

ericcornelissen (changes requested):

public/stylesheet.css
ericcornelissen: Hmm :thinking: seeing a similar problem at a slightly larger size (this is exactly 682px wide):

image

tests/e2e.test.js
jorgeamadosoria: fixed.

public/stylesheet.css
ericcornelissen: @jorgeamadosoria, did you see this comment?

jorgeamadosoria: no. I just got notified now.

jorgeamadosoria: fixed

public/index.pug
```diff
@@ -175,5 +174,7 @@ html(lang="en-US")
...
+            #[a(title="github repository" href="https://github.com/simple-icons/simple-icons-website") Made with &#10084; on GitHub]
```
ericcornelissen: We should probably distinguish this `title` from [this one](https://github.com/simple-icons/simple-icons-website/blob/7843220e0a2b825dffc2e3f58fb9d41fae7e0e5b/public/index.pug#L32) more clearly, here's some suggestions:

- header: "project repository", "main repository", "icon repository"
- footer: "website repository", "website source code", "website source"

ericcornelissen (commented):
This looks pretty good, got one minor thing left that we should probably take care of before merging this

public/index.pug
jorgeamadosoria: true. Done.

ericcornelissen (approved):
Cool, thanks for this @jorgeamadosoria :+1: 

Also, great to have those e2e tests fixed :tada:

Issue #17, opened 2021-01-11
Add compact layout


ericcornelissen:
**:information_source: If you're interested in this feature, please leave comment or add a :+1:**

---

### Kind of issue 

- [x] Feature
- [ ] Bug
- [x] Design
- [ ] Other, namely:

### Description

Following [the original design](https://github.com/simple-icons/simple-icons/issues/980#issuecomment-427185048), a "compact layout" mode can be added to the website. The end-design may differ from what is displayed there, it is just a guideline.

jorgeamadosoria:
Are there more details on this? Is this issue still relevant?

ericcornelissen:
No more details than in the comment that is linked. Essentially it would entail adding an option to use a more "compact" design. I.e. the cards would be smaller, more cards could fit on the screen at once, but touch targets would be smaller. This would be similar to something that Google has in various apps (including Gmail) where you can choose between a "Comfortable" design and a "Compact" design.

It is relevant in the sense that there's interest (currently 5 :+1:s) and it hasn't been added yet.

jorgeamadosoria:
I was think that the implementation of the. menu with icons, the universal download type, and the icon detail issues would diminish the relevance of this.

The menu is now shorter, the universal type would make the card way shorter , the detail view would allow many options for each icon without overloading the page....
I think we should reassess how this would look after addressing those other issues? Because it would certainly look different.

Not discarding the relevance of the issue, but rather asking for a more detail view on how this would look after implementing those other things.

ericcornelissen:
At least to me, all the points you raised are almost irrelevant to this issue. First off, the compact design is mostly targeted towards the cards/grid items, so the "menu" is by-and-large out of scope. Second, a key difference between a "Compact" and "Comfortable" design is the size of elements and the amount margins/paddings in a design. In a "Compact" design things are closer together and click targets are smaller. I think this still applies after addressing #33 and even #43 (in my eyes the detail view should only add info currently not displayed, not move info currently displayed into the details view).

jorgeamadosoria:
> (in my eyes the detail view should only add info currently not displayed, not move info currently displayed into the details view).

Ah, but that doesn't has to be the case. Info that is currently shown but doesn't have to be include license and brand guidelines, for example.

Also, implementing the universal download type removes one download option from the card, which can help to redesign it even in the current layout. That will affect the compact layout as well. 
Thus, it is important to have clarity on what the compact layout looks like, preferably after implementing the other ones.

[...]

PR #112, opened 2021-09-12
Added a group of buttons for universal download type

### LINKED ISSUE #33: [Suggestion] Universally choose download type

---

### Kind of issue 

- [x] Improvement
- [ ] Bug
- [ ] Other, namely:

### Description

Following an idea that appeared in [this redesign proposal](https://github.com/simple-icons/simple-icons/issues/980#issuecomment-547591284) we may consider updating the website so that the user can choose what kind of files they want to download (currently SVG or PDF) and than have single download button in grid items to download the brand logo in that format.

This has two potential benefits:

- If we provide support for more formats (e.g. PNG and JPEG), the buttons may no longer fit in the grid item, requiring us to redesign the grid items to accommodate many buttons.
- If there is just one button to download the icon, it is less likely the wrong button is pressed.

It also has some potential benefits

- As this is maybe not super common, the user may not find out that they can change the type of file that will be downloaded. Which is a shame if they're not looking for the default.
- It may be non-trivial to support this for users without JavaScript. Though, we may choose to generate one page for downloading SVGs and one for PDFs (and one for whatever we wish to support) and change the button to choose the type into a link to that page.

This issue relates to #25 as it would probably lead to a redesign of the grid items, since multiple download buttons are no longer needed.

jorgeamadosoria:
issue: #33 

image

image

The button selected determines which file is downloaded from the arrow icon.

The code requires refactoring to avoid duplication in the link.


jorgeamadosoria:
Can we get some feedback on this @simple-icons/maintainers @ericcornelissen ?

The approval / rejection of this PR is blocking me to move on to #17

ericcornelissen:
Did you try to make the behaviour of the download button dynamic? That is, rather than having two buttons and switching between them, can you change the behaviour of the button based on the selected type?

As an aside, I'm personally not really a fan of the Acrobat logo to denote PDF but don't necessarily have a good suggestion either.

jorgeamadosoria:
> Did you try to make the behaviour of the download button dynamic? That is, rather than having two buttons and switching between them, can you change the behaviour of the button based on the selected type?
> 
> As an aside, I'm personally not really a fan of the Acrobat logo to denote PDF but don't necessarily have a good suggestion either.

Haven't tried any other approaches, I meant this PR to be a starting point to get recommendations.

I'm also not big on the adobe logo because that's ADOBE, not the PDF format. I used it because, well, Adobe Reader has become attached to the PDF format in common web parlance (however incorrect that is) and it's a logo we have in simple icons, so it requires zero dependencies or size or a lot of code. So, it's convenient, mostly, that's really the reason to use it.

ericcornelissen:
> Haven't tried any other approaches, I meant this PR to be a starting point to get recommendations.

What do you think about the approach I suggested?


> I'm also not big on the adobe logo because that's ADOBE, not the PDF format. I used it because, well, Adobe Reader has become attached to the PDF format in common web parlance (however incorrect that is) and it's a logo we have in simple icons, so it requires zero dependencies or size or a lot of code. So, it's convenient, mostly, that's really the reason to use it.

I see, I'd personally prefer to have a different logo (embedded directly as an SVG like the other buttons) if the only concern is convenience. But, as I said in my previous comment, I don't have a suggestion for an alternative either.

jorgeamadosoria:
> What do you think about the approach I suggested?
> 

I like it, it's the most logical things to do. I'm going to try tonight.

> 
> I see, I'd personally prefer to have a different logo (embedded directly as an SVG like the other buttons) if the only concern is convenience. But, as I said in my previous comment, I don't have a suggestion for an alternative either.

Well, there is also that, we would need a suitable replacement. Alternatively, I could just put a PDF / SVG title in bold letters nad that's it, I didn't do it because every other button has an icon.

jorgeamadosoria:
> > What do you think about the approach I suggested?
> 
> I like it, it's the most logical things to do. I'm going to try tonight.
> 
> > I see, I'd personally prefer to have a different logo (embedded directly as an SVG like the other buttons) if the only concern is convenience. But, as I said in my previous comment, I don't have a suggestion for an alternative either.
> 
> Well, there is also that, we would need a suitable replacement. Alternatively, I could just put a PDF / SVG title in bold letters nad that's it, I didn't do it because every other button has an icon.

@ericcornelissen I've implemented a change as you suggested. Pending issues I see:
- A different identification for the buttons. Maybe just text?
- Tests.

I had to implement a different  handler for the download arrow svg because it wasn't downloading if the user clicked not the link but explicitly the arrow. The handler simply triggers the link handler, so it's not new code.

Unrelated, but to reduce the size of the page, can't we remove some svg attributes like role, viewBox or the namespace? the svgs seem not to need those to show in the website, and those contribute their size in bytes multiplied by the amount of svgs, which is ever increasing.

jorgeamadosoria:
alternative svgs for the download buttons:

image

I even have one ready for PNG when I add those later

tests/e2e.test.js
```diff
@@ -418,12 +418,13 @@ describe('Grid item', () => {
...
-  });
+  it('is possible to download an icon in all existing file types', async () => {
+    const fileTypes = ['download-svg', 'download-pdf'];
 
-  it('is possible to download an icon as PDF', async () => {
-    await expect(page).toClick('a[download][href$="pdf"]');
+    fileTypes.forEach(async (fileType) => {
+      await expect(page).toClick(`button#${fileType}`);
+      await expect(page).toClick(`a[download].download-file`);
+    });
```
ericcornelissen: Could you rewrite using `it.each`, that would make it a bit easier to tell which download option isn't working if the test fails.

public/index.pug
```diff
@@ -99,6 +99,15 @@ html(lang="en-US")
...
                 path(d="M12 4a1 1 0 0 1-1-1V1a1 1 0 0 1 2 0v2a1 1 0 0 1-1 1zM4.933 6.348a.997.997 0 0 0 1.414 0 .999.999 0 0 0 0-1.414l-1.42-1.42a.999.999 0 1 0-1.414 1.414l1.42 1.42zM1 13h2a1 1 0 1 0 0-2H1a1 1 0 0 0 0 2zm19.486-8.072l3.221-3.221A.999.999 0 1 0 22.293.293l-3.221 3.221-1.42 1.42-2.19 2.19A5.955 5.955 0 0 0 12 6c-3.309 0-6 2.691-6 6 0 1.258.406 2.453 1.124 3.462l-2.105 2.105c-.026.021-.058.03-.083.055s-.033.056-.055.082l-1.368 1.368-.001.002-3.219 3.219a.999.999 0 1 0 1.414 1.414l3.987-3.987a10.03 10.03 0 0 0 6.332 2.262c5.103-.001 9.473-3.902 9.951-9.081a1 1 0 0 0-1.591-.896 5.96 5.96 0 0 1-7.037.06l5.717-5.716 1.42-1.421zm-.945 9.78c-1.21 3.337-4.564 5.587-8.257 5.238a8.019 8.019 0 0 1-4.165-1.651l4.802-4.802c.05.038.093.082.144.12a7.955 7.955 0 0 0 7.476 1.095zm-10.979-.684A3.968 3.968 0 0 1 8 12c0-2.206 1.794-4 4-4a3.98 3.98 0 0 1 2.025.561l-5.463 5.463z")
+        div(class="control-field")
+          label(class="control-field__title") Download Type
+          div(class="control-field__row")
+            button(id="download-pdf" class="control__button" title="download PDF" disabled)
```
ericcornelissen: Can we update the ID for clarity, while reviewing the JavaScript I thought this ID referred to something that is for downloading a PDF.

```suggestion
            button(id="download-type-pdf" class="control__button" title="download PDF" disabled)
```

public/index.pug
```diff
@@ -99,6 +99,15 @@ html(lang="en-US")
...
+            button(id="download-svg" class="control__button" title="download SVG" disabled)
```
ericcornelissen: Can we update the ID for clarity, while reviewing the JavaScript I thought this ID referred to something that is for downloading an SVG.

```suggestion
            button(id="download-type-svg" class="control__button" title="download SVG" disabled)
```

public/scripts/download-type.js
```diff
@@ -0,0 +1,68 @@
...
+    selectDownloadType(SVG_DOWNLOAD_TYPE);
+    $downloadSvg.blur();
+  });
+
+  for (var i = 0; i < $downloadFiles.length; i++) {
+    // svg event handler
+    $downloadFiles[i].children[0].addEventListener('click', (event) => {
+      event.preventDefault();
+      event.target.closest('a').click();
+    });
```
ericcornelissen: > I had to implement a different handler for the download arrow svg because it wasn't downloading if the user clicked not the link but explicitly the arrow. The handler simply triggers the link handler, so it's not new code.

Does that comment refer to this snippet of code? If so, you can simply add the CSS rule[`pointer-events: none;`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) to the SVG.

public/stylesheet.css
```diff
@@ -474,8 +475,20 @@ body.order-by-relevance #order-relevance {
   fill: var(--color-background);
 }
 
-/* Control - color */
```
ericcornelissen, line 549: Minor, for consistency with the rest of the code you could add here

```css
/* Control - download type */
```

public/index.pug
```diff
@@ -99,6 +99,15 @@ html(lang="en-US")
...
               svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" aria-labelledby="color-scheme-system")
                 path(d="M12 4a1 1 0 0 1-1-1V1a1 1 0 0 1 2 0v2a1 1 0 0 1-1 1zM4.933 6.348a.997.997 0 0 0 1.414 0 .999.999 0 0 0 0-1.414l-1.42-1.42a.999.999 0 1 0-1.414 1.414l1.42 1.42zM1 13h2a1 1 0 1 0 0-2H1a1 1 0 0 0 0 2zm19.486-8.072l3.221-3.221A.999.999 0 1 0 22.293.293l-3.221 3.221-1.42 1.42-2.19 2.19A5.955 5.955 0 0 0 12 6c-3.309 0-6 2.691-6 6 0 1.258.406 2.453 1.124 3.462l-2.105 2.105c-.026.021-.058.03-.083.055s-.033.056-.055.082l-1.368 1.368-.001.002-3.219 3.219a.999.999 0 1 0 1.414 1.414l3.987-3.987a10.03 10.03 0 0 0 6.332 2.262c5.103-.001 9.473-3.902 9.951-9.081a1 1 0 0 0-1.591-.896 5.96 5.96 0 0 1-7.037.06l5.717-5.716 1.42-1.421zm-.945 9.78c-1.21 3.337-4.564 5.587-8.257 5.238a8.019 8.019 0 0 1-4.165-1.651l4.802-4.802c.05.038.093.082.144.12a7.955 7.955 0 0 0 7.476 1.095zm-10.979-.684A3.968 3.968 0 0 1 8 12c0-2.206 1.794-4 4-4a3.98 3.98 0 0 1 2.025.561l-5.463 5.463z")
+        div(class="control-field")
+          label(class="control-field__title") Download Type
```
ericcornelissen: Wondering if we can get away here with just "Download" :thinking: 

```suggestion
          label(class="control-field__title") Download
```

ericcornelissen (commented):
This is looking really good! 

> Pending issues I see: Tests.

Do you think you can add unit tests based on existing tests yourself?

> alternative svgs for the download buttons:

I do quite like the approach, though I think they're a bit too small at the moment.

> Unrelated, but to reduce the size of the page, can't we remove some svg attributes like role, viewBox or the namespace? the svgs seem not to need those to show in the website, and those contribute their size in bytes multiplied by the amount of svgs, which is ever increasing.

Feel free to open a new issue to discuss this.

tests/e2e.test.js
jorgeamadosoria: I'm not familiar enough with jest. I will rewrite it as you said.

public/index.pug
jorgeamadosoria: sure. I won't commit the suggestion because this id is used elsewhere to identify this button

public/index.pug
jorgeamadosoria: Sure.

public/scripts/download-type.js
jorgeamadosoria: nice! I'll try that.

public/stylesheet.css
jorgeamadosoria: sure

public/index.pug
jorgeamadosoria: not sure. It may be read as "these buttons will download" and not "these buttons will signal what to download" as it is the only reading now.

jorgeamadosoria: It may work too. I honestly don't know. we may need a third opinion here. @sachinraja ?

jorgeamadosoria:
> This is looking really good!
> 
> > Pending issues I see: Tests.
> 
> Do you think you can add unit tests based on existing tests yourself?

Yes, I can try. 

> > alternative svgs for the download buttons:
> 
> I do quite like the approach, though I think they're a bit too small at the moment.

I've increased the size of the text to 34px. more than that and the text would touch the button border, which creates a visually unpleasant effect when changing color scheme and in the future when the buttons are joined. I think that's the maximum size the letters can have now:

image

public/index.pug
```diff
@@ -143,17 +152,14 @@ html(lang="en-US")
...
-                href=`./icons/${icon.slug}.pdf`
+                class="download-type-file grid-item__download-button grid-item__button"
+                href='#'
                 role="button"
+                data-icon=`${icon.slug}`
                 download
-              ) PDF
+              )
+                svg(height="12" width="12" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" aria-labelledby="download")
+                  path(d="m11.2 0c-0.442 0-0.8 0.358-0.8 0.8v11.4l-3.14-2.76c-0.331-0.29-0.835-0.258-1.13 0.0735l-1.05 1.2c-0.293 0.333-0.259 0.84 0.0735 1.13l6.33 5.54c0.302 0.264 0.752 0.264 1.05 0l6.32-5.54c0.333-0.291 0.366-0.798 0.0735-1.13l-1.05-1.2c-0.291-0.331-0.795-0.363-1.13-0.0735l-3.14 2.76v-11.4c0-0.442-0.358-0.8-0.8-0.8zm-8 20.8c-0.442 0-0.8 0.358-0.8 0.8v1.6c0 0.442 0.358 0.8 0.8 0.8h17.6c0.442 0 0.8-0.358 0.8-0.8v-1.6c0-0.442-0.358-0.8-0.8-0.8z")
```
ericcornelissen: > Unrelated, but to reduce the size of the page, can't we remove some svg attributes like role, viewBox or the namespace? the svgs seem not to need those to show in the website, and those contribute their size in bytes multiplied by the amount of svgs, which is ever increasing.

Coming back to this comment (tangentially) w.r.t. this specific SVG, we could use an approach similar to what we [used to do for the Twitter icon in the footer](https://github.com/simple-icons/simple-icons-website/blob/18dec0364394a246bca13b1ba1e37689ef7df079/public/index.pug#L157), i.e. have the SVG once and use the `<use>` tag to refer it everywhere else.

My gut feeling is that it's probably easiest to have the SVG once outside the list (hidden) and `<use>` it throughout the list.

tests/e2e.test.js
```diff
@@ -417,13 +417,22 @@ describe('Grid item', () => {
...
-    await expect(page).toClick('a[download][href$="svg"]');
+    fileTypes.forEach(async (fileType) => {
+      await expect(page).toClick(`button#${fileType}`);
+      await expect(page).toClick(`a[download].download-type-file`);
+    });
   });
-
-  it('is possible to download an icon as PDF', async () => {
-    await expect(page).toClick('a[download][href$="pdf"]');
+*/
```
ericcornelissen: Is this a leftover?

tests/e2e.test.js
```diff
@@ -417,13 +417,22 @@ describe('Grid item', () => {
...
+  it.each([
+    ['download-type-svg', 'svg'],
+    ['download-type-pdf', 'pdf'],
+  ])('is possible to download an icon in %s file type', async (fileType) => {
```
ericcornelissen: I don't think this works as you want it to work, you'd need to do:

```suggestion
  it.each([
    ['svg', 'download-type-svg'],
    ['pdf', 'download-type-pdf'],
  ])('is possible to download an icon in %s file type', async (_, fileType) => {
```

But, at least to me, something like the following is clear enough:

```suggestion
  it.each([
    'download-type-svg',
    'download-type-pdf',
  ])('is possible to download an icon (%s)', async (fileType) => {
```

public/scripts/download-type.js
```diff
@@ -0,0 +1,62 @@
...
+    selectDownloadType(PDF_DOWNLOAD_TYPE);
+    $downloadPdf.blur();
+  });
+  $downloadSvg.addEventListener('click', (event) => {
+    event.preventDefault();
+    selectDownloadType(SVG_DOWNLOAD_TYPE);
+    $downloadSvg.blur();
+  });
+
+  for (var i = 0; i < $downloadFiles.length; i++) {
```
ericcornelissen: I think we should prefer `let` over `var`

```suggestion
  for (let i = 0; i < $downloadFiles.length; i++) {
```

Or, alternatively, in line with the other code, we could write it as

```
$downloadFiles.forEach(($downloadFile) => {
  // add click listener
});

File: public/index.pug [hunk 11]
```diff
@@ -143,17 +152,14 @@ html(lang="en-US")
...
-                class="grid-item__button"
-                href=`./icons/${icon.slug}.svg`
-                role="button"
-                download
-              ) SVG
-              a(
-                class="grid-item__button"
-                href=`./icons/${icon.slug}.pdf`
+                class="download-type-file grid-item__download-button grid-item__button"
+                href='#'
```
ericcornelissen: Should (can?) we set this to the download URL for SVGs by default?

jorgeamadosoria:
I don't see how to use the <use> tag. it's surprisingly difficult to find documentation for that tag, and in the example linked I couldn't find the svg definition, so I don't really know what to do there.

jorgeamadosoria:
I have uploaded a sprite.svg file to address the download svg repetition. I've also moved all other website svgs to that sprite file.

jorgeamadosoria:
I have uploaded a sprite.svg file to address the download svg repetition. I've also moved all other website svgs to that sprite file.

tests/e2e.test.js
```diff
@@ -426,13 +426,13 @@ describe('Grid item', () => {
...
-  });
-
-  it('is possible to download an icon as PDF', async () => {
-    await expect(page).toClick('a[download][href$="pdf"]');
-  });
+  it.each(['download-type-svg', 'download-type-pdf'])(
+    'is possible to download an icon "%s"',
+    async (fileType) => {
+      await expect(page).toClick(`button#${fileType}`);
+      await expect(page).toClick(`a[download].download-type-file`);
```
ericcornelissen: Any particular reason you're using backticks here?

```suggestion
      await expect(page).toClick('a[download].download-type-file');
```

public/stylesheet.css
```diff
@@ -474,8 +478,24 @@ body.order-by-relevance #order-relevance {
...
 
+.grid-item__download-button:hover,
+.grid-item__download-button:focus {
+  fill: var(--color-background);
+  background-color: var(--color-button-active);
+}
+
+.download-type-file svg {
+  pointer-events: none;
+}
```
ericcornelissen: This actually seems to work fine for me in both Firefox and Chrome if I remove it :thinking:

public/images/sprite.svg
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ericcornelissen: This filename doesn't seem very descriptive to me

public/stylesheet.css
```diff
@@ -402,6 +402,11 @@ body.no-js .control-field__title {
   margin: 0 0.3rem;
   width: 2.4rem;
 }
+
+.control__button__download {
+  padding: 1px;
+}
```
ericcornelissen: If at all possible I'd prefer it if we could avoid this class. It seems to me overly specific, and a padding of 1 pixel for this use case does not seem worth it to me. You should be able to modify the SVGs such that the display as you expect, similar to the other SVGs in the control section.

If you can't @jorgeamadosoria, feel free to leave it

public/index.pug
```diff
@@ -143,17 +152,15 @@ html(lang="en-US")
                 disabled
               ) ##{icon.hex}
               a(
-                class="grid-item__button"
+                class="download-type-file grid-item__download-button grid-item__button"
```
ericcornelissen: Whats the point of having both `grid-item__download-button` and `grid-item__button`, both of which only appear on this element?

public/stylesheet.css
```diff
@@ -474,8 +478,24 @@ body.order-by-relevance #order-relevance {
...
+.grid-item__download-button {
+  display: inline-block;
+  fill: var(--color-button-active);
+}
 
+.grid-item__download-button:hover,
+.grid-item__download-button:focus {
+  fill: var(--color-background);
+  background-color: var(--color-button-active);
+}
```
ericcornelissen: Related to the previous comment, why are these added if there's already styling for `.grid-item__button`?

jorgeamadosoria:
Why can't I respond to your comments in a thread?

public/stylesheet.css
ericcornelissen: > Why can't I respond to your comments in a thread?

@jorgeamadosoria, no idea but I still can. It could be that GitHub thinks you're doing a review. You could try cleaning your browser history/state for the GitHub domain (or try incognito first to check).

jorgeamadosoria:
Apparently, using the `use` tag removed the need for the events:none attribute (and the need for that class altogether). Removed the download-type-file class.

jorgeamadosoria:
> If at all possible I'd prefer it if we could avoid this class. It seems to me overly specific, and a padding of 1 pixel for this use case does not seem worth it to me. You should be able to modify the SVGs such that the display as you expect, similar to the other SVGs in the control section.

The problem here is that all control__button buttons have a padding set to 0.4rem. I can style the SVG but it will always be constrained by the button padding unless I overwrite it, which is what's happening here.
I see several options: 
- Don't have control__button classes for the download control buttons. I don't like this for consistency reasons.
- Refactor all SVGs to take all the space in the button, and remove the padding completely. I don't like it because it changes perfectly fine buttons due to the addition of new buttons.
- Separate padding in a different class, and apply that differently to each button. 
- Leave it as is.

I'm fine with either of the last two. @ericcornelissen let me know if you prefer one.

jorgeamadosoria:
> This filename doesn't seem very descriptive to me

sprite.svg is straightforward. It's a sprite, and it's an svg. I you have any other preferences let me know, I can change it. but as it stands, I wouldn't know how else to call it.

jorgeamadosoria:
I've addressed all the comments except the sprite name.

I added two separate padding classes to the control buttons as an alternative to CSS overwriting the padding attribute.

public/stylesheet.css
```diff
@@ -475,7 +482,10 @@ body.order-by-relevance #order-relevance {
 }
 
 <<<<<<< Updated upstream
+<<<<<<< Updated upstream
+=======
 =======
+>>>>>>> Stashed changes
 
 >>>>>>> Stashed changes
```
sachinraja: These should definitely be removed

public/stylesheet.css
```diff
@@ -378,6 +378,13 @@ body.no-js .control-field__title {
   border-radius: var(--border-radius-small);
   color: var(--color-text-default);
   height: 2.4rem;
+}
+
+.padding_min {
+  padding: 1px;
+}
+
+.padding_large {
```
ericcornelissen: "min" and "large" are not opposites of each other, I'd prefer something more intuitive like "small" and "large" or "min" and "max".

ericcornelissen: > I added two separate padding classes to the control buttons as an alternative to CSS overwriting the padding attribute.

It's your call on the approach, but I think it's best to just have all the control buttons have the same padding.

jorgeamadosoria:
> It's your call on the approach, but I think it's best to just have all the control buttons have the same padding.

@ericcornelissen, we can't have a large PDF / SVG lettering AND keep the same padding. We've tried:
- Adobe and SVG icons with the same padding as the other buttons: Discarded because the Adobe icon shouldn't represent PDFs
- PDF / SVG lettering with the same padding as the other buttons: Discarded because the lettering was too small.
- PDF / SVG lettering with minimum padding (1px): In discussion because the padding is not the same as the other buttons.

I'm good with any of these choices. Take your pick or suggest a different one, I'll go with it. For me this is a very small point I'll be happy to move on from as soon as you are happy with the choice made.

jorgeamadosoria:
the remaining items I see now are:
- Make a decision on the download buttons design.
- The `use` tag in the download icon makes the arrow disappear when clicked. If I remove the `use` tag that doesn't happen. I'm not sure if this can be fixed with CSS classes or the only way to fix it would be to forget about the sprite and the `use` tag. I need help with this.

public/index.pug
```diff
@@ -25,37 +25,37 @@ html(lang="en-US")
...
 
-  body(class="no-js order-alphabetically")
+  body(class="no-js order-alphabetically download-svg")
     header(class="header")
       ul(class="header__list")
         li(class="header__list-item")
           a(class="header__list-icon" title="Main Repository" href="https://github.com/simple-icons/simple-icons")
-            svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
-              path(d="M12 .297c-6.63 0-12 5.373-12 12 0 5.303 3.438 9.8 8.205 11.385.6.113.82-.258.82-.577 0-.285-.01-1.04-.015-2.04-3.338.724-4.042-1.61-4.042-1.61C4.422 18.07 3.633 17.7 3.633 17.7c-1.087-.744.084-.729.084-.729 1.205.084 1.838 1.236 1.838 1.236 1.07 1.835 2.809 1.305 3.495.998.108-.776.417-1.305.76-1.605-2.665-.3-5.466-1.332-5.466-5.93 0-1.31.465-2.38 1.235-3.22-.135-.303-.54-1.523.105-3.176 0 0 1.005-.322 3.3 1.23.96-.267 1.98-.399 3-.405 1.02.006 2.04.138 3 .405 2.28-1.552 3.285-1.23 3.285-1.23.645 1.653.24 2.873.12 3.176.765.84 1.23 1.91 1.23 3.22 0 4.61-2.805 5.625-5.475 5.92.42.36.81 1.096.81 2.22 0 1.606-.015 2.896-.015 3.286 0 .315.21.69.825.57C20.565 22.092 24 17.592 24 12.297c0-6.627-5.373-12-12-12")
+            svg(viewBox="0 0 24 24")
```
ericcornelissen: I'm guessing the motivation for removing these attributes is to reduce the size of the website. Are you 100% this is safe?

```suggestion
            svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
```

(same for all the other SVGs)

**Tip**: If you want this PR to be merged quicker, move this change to another PR and this won't be blocked by this discussion.

public/index.pug
```diff
@@ -25,37 +25,37 @@ html(lang="en-US")
...
+              use(href="./images/sprite.svg#menu-main-repository")
         li(class="header__list-item")
           a(class="header__list-icon" title="npm" href="https://www.npmjs.com/package/simple-icons")
-            svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg")
-              path(d="M1.763 0C.786 0 0 .786 0 1.763v20.474C0 23.214.786 24 1.763 24h20.474c.977 0 1.763-.786 1.763-1.763V1.763C24 .786 23.214 0 22.237 0zM5.13 5.323l13.837.019-.009 13.836h-3.464l.01-10.382h-3.456L12.04 19.17H5.113z")
+            svg(viewBox="0 0 24 24")
+              use(href="./images/sprite.svg#menu-npm")
```
ericcornelissen: I'd like to invite all @simple-icons/maintainers and anyone else to weigh into the topic of using sprites here. It seems fine to me but I'm not too familiar with using sprites for this purpose so if anyone with more experience has any feedback that'd be very welcome :slightly_smiling_face:

public/stylesheet.css
```diff
@@ -634,21 +637,23 @@ body:not(.dark):not(.light) #color-scheme-system:not(:disabled) {
...
 .grid-item__button {
+  fill: var(--color-button-active);
   border-top: 1px solid var(--color-grid-item-divider);
   background-color: var(--color-background);
-  color: var(--color-button-active);
   flex-grow: 1;
   font-family: var(--font-family-stylized);
   font-size: 0.75rem;
   font-weight: 600;
-  outline: none;
```
ericcornelissen: Did you remove this intentionally?

public/stylesheet.css
```diff
@@ -634,21 +637,23 @@ body:not(.dark):not(.light) #color-scheme-system:not(:disabled) {
...
   text-align: center;
 }
+
 .grid-item__button:focus,
 .grid-item__button:hover {
+  fill: var(--color-background);
+  background-color: var(--color-button-active);
   background: var(--color-button-hover);
-  color: var(--color-button-text-hover);
+  fill: var(--color-button-text-hover);
```
ericcornelissen: Two times `fill` and two times `background`?


```css
background: var(--color-button-hover);
fill: var(--color-button-text-hover);
```

webpack.config.js
```diff
@@ -16,6 +16,28 @@ const NODE_MODULES = path.resolve(__dirname, 'node_modules');
...
+      const module = line.split(' | ')[0];
+      const author = line.split(' | ')[1];
+      return {
+        nameModule: module.match(/\[(.*?)\]/gs)[0].slice(1, -1),
+        urlModule: module.match(/\((.*?)\)/gs)[0].slice(1, -1),
+        nameAuthor: author.match(/\[(.*?)\]/gs)[0].slice(1, -1),
+        urlAuthor: author.match(/\((.*?)\)/gs)[0].slice(1, -1),
+      };
+    });
+}
```
ericcornelissen: Wrong PR?

public/scripts/download-type.js
```diff
@@ -0,0 +1,62 @@
...
+
+  if (storage.hasItem(STORAGE_KEY_DOWNLOAD_TYPE)) {
+    const storedDownloadType = storage.getItem(STORAGE_KEY_DOWNLOAD_TYPE);
+    selectDownloadType(storedDownloadType);
+  }
+
+  $downloadPdf.addEventListener('click', (event) => {
+    event.preventDefault();
+    selectDownloadType(PDF_DOWNLOAD_TYPE);
+    $downloadPdf.blur();
```
ericcornelissen: Following #126 can you remove this?

```suggestion
```

public/scripts/download-type.js
```diff
@@ -0,0 +1,62 @@
...
+  });
+  $downloadSvg.addEventListener('click', (event) => {
+    event.preventDefault();
+    selectDownloadType(SVG_DOWNLOAD_TYPE);
+    $downloadSvg.blur();
```
ericcornelissen: Following #126 can you remove this?

```suggestion
```

public/index.pug
```diff
@@ -78,27 +78,36 @@ html(lang="en-US")
...
+              svg(viewBox="0 0 24 24" aria-labelledby="color-scheme-dark")
+                use(href="./images/sprite.svg#color-scheme-dark")
+            button(id="color-scheme-system" class="control__button__padding control__button last__button" title="System color scheme" disabled)
+              svg(viewBox="0 0 24 24" aria-labelledby="color-scheme-system")
+                use(href="./images/sprite.svg#color-scheme-system")
+        div(class="control-field")
+          label(class="control-field__title") Download
+          div(class="control-field__row")
+            button(id="download-pdf" class="control__button first__button" title="download PDF" disabled)
+              svg(height="34" viewBox="0 0 24 24" aria-labelledby="download-svg")
```
ericcornelissen: Using the `height` attribute here seems like the wrong approach to styling this element. Could you find an alternative solution? Ideally one where you don't need to style the SVG to get the desired size of the SVG (but rather rely on e.g. padding of the container) OR apply the styling in CSS.

ericcornelissen (changes requested):

jorgeamadosoria:
moved the svg(height="34"...) inline styling to the equivalent CSS style. height=34px is good for the button taking padding into consideration.

jorgeamadosoria:
fixed the double fill / background on .grid__button hover

public/index.pug
ericcornelissen: @jorgeamadosoria

ericcornelissen:
> moved the svg(height="34"...) inline styling to the equivalent CSS style. height=34px is good for the button taking padding into consideration.

<details>
<summary>Can you elaborate? Why do we have to force the height to a specific amount of pixels? Specifying the exact pixel value is not very dynamic and is hard to keep consistent across the whole project.</summary>

I'll admit that the existing CSS is not perfect in this regard, but that's not a reason to add more bad CSS to the pile.
</details>

Why not have it fit the container it's in, like the other buttons. If you can explain WHY you need to specify the exact pixel value then feel free to leave it as is.

webpack.config.js
```diff
@@ -4,8 +4,8 @@ const getRelativeLuminance = require('get-relative-luminance').default;
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const path = require('path');
+const fs = require('fs');
 const simpleIcons = require('simple-icons');
-
```
ericcornelissen: Wrong PR?

public/stylesheet.css
ericcornelissen: @jorgeamadosoria

jorgeamadosoria:
Removed sprite, removed fixed height

jorgeamadosoria:
refactored the download arrow in the card footer

public/stylesheet.css
ericcornelissen: @jorgeamadosoria, it appears this change is still here. If you have a reason to remove it, please share it. If not, please put it back.

public/stylesheet.css
```diff
@@ -634,21 +639,26 @@ body:not(.dark):not(.light) #color-scheme-system:not(:disabled) {
...
+  height: 1rem;
+  pointer-events: none;
+}
+
 .grid-item__button:focus,
 .grid-item__button:hover {
-  background: var(--color-button-hover);
-  color: var(--color-button-text-hover);
+  fill: var(--color-background);
+  background-color: var(--color-button-active);
```
ericcornelissen: Did you intentionally change the background color from `var(--color-button-hover)` to `var(--color-button-active)`?

public/stylesheet.css
```diff
@@ -634,21 +639,26 @@ body:not(.dark):not(.light) #color-scheme-system:not(:disabled) {
...
+.grid-item__button svg {
+  height: 1rem;
+  pointer-events: none;
+}
+
 .grid-item__button:focus,
 .grid-item__button:hover {
-  background: var(--color-button-hover);
-  color: var(--color-button-text-hover);
+  fill: var(--color-background);
```
ericcornelissen: Did you intentionally change the color of "the content" (whether text or an SVG) from `var(--color-button-text-hover)` to `var(--color-background)`. I guess it looks fine but "color background" seem like a non-ideal choice for something that is not the background.

public/stylesheet.css
```diff
@@ -615,6 +619,7 @@ body:not(.dark):not(.light) #color-scheme-system:not(:disabled) {
   flex-wrap: nowrap;
   justify-content: flex-start;
   margin-top: 0.4rem;
+  height: 2rem;
```
ericcornelissen: Where did this change come from? Seems unrelated to the rest of the PR.

public/stylesheet.css
```diff
@@ -407,6 +410,7 @@ body.no-js .control-field__title {
   border-left-width: 0.1rem;
   border-top-left-radius: var(--border-radius-small);
   border-bottom-left-radius: var(--border-radius-small);
+  width: 2.5rem;
```
ericcornelissen: Where did this change come from? Seems unrelated to the rest of the PR.

public/stylesheet.css
```diff
@@ -634,21 +639,26 @@ body:not(.dark):not(.light) #color-scheme-system:not(:disabled) {
...
   font-size: 0.75rem;
   font-weight: 600;
-  outline: none;
   padding: 8px;
   text-align: center;
 }
+
+.grid-item__button svg {
+  height: 1rem;
+  pointer-events: none;
```
ericcornelissen: Didn't we agree this isn't needed? Did you put it back intentionally?

public/stylesheet.css
```diff
@@ -376,6 +376,9 @@ body.no-js .control-field__title {
   background-color: transparent;
   color: var(--color-text-default);
   height: 2.4rem;
+}
+
+.control__button__padding {
```
ericcornelissen: Why not add this padding so the buttons being added here look (arguably) better?

| Current | With padding |
| :---: | :---: |
| current | with padding |

I know we've been going back and forth about this, any objections to this suggestion?

jorgeamadosoria:
image

I remove it because it doesn't seem to do anything.

public/index.pug
```diff
@@ -146,14 +155,12 @@ html(lang="en-US")
...
                 download
-              ) SVG
-              a(
-                class="grid-item__button"
-                href=`./icons/${icon.slug}.pdf`
-                role="button"
-                download
-              ) PDF
+              )
+                svg(role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" aria-labelledby="download")
```
ericcornelissen: The `aria-labelledby` value should reference an element with the `id="download"` that either has text content or a `title` attribute value. In it's current state (without such an element) it does nothing.

public/index.pug
```diff
@@ -146,14 +155,12 @@ html(lang="en-US")
                 class="grid-item__button"
                 href=`./icons/${icon.slug}.svg`
                 role="button"
+                data-icon=`${icon.slug}`
```
ericcornelissen: You can actually do

```suggestion
                data-icon=icon.slug
```

which makes your intend a bit clearer (namely that the `data-icon` value should be exactly and only the icon's slug).

ericcornelissen (changes requested):

public/stylesheet.css
ericcornelissen: > image
> 
> I remove it because it doesn't seem to do anything.

Your UI says your comment is **Pending**, which means no one else can see it

That said, it does affect rendering. All (major) browsers by default add `outline`s to tabable elements as a visual indicator of what element on the page is currently in focus. As we have our own styling for that we remove the outline (a decision open to discussion, but ideally we don't change that until a discussion as taken place AND consistently across the website in one go).

public/index.pug
```diff
@@ -78,27 +78,36 @@ html(lang="en-US")
...
+        div(class="control-field")
+          label(class="control-field__title") Download
+          div(class="control-field__row")
+            button(id="download-pdf" class="control__button first__button" title="download PDF" disabled)
+              svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" aria-labelledby="download-svg")
```
mondeja:
```suggestion
              svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" aria-labelledby="download-pdf")
```

public/index.pug
```diff
@@ -78,27 +78,36 @@ html(lang="en-US")
...
...9e-8 l 0.0062,-4.661004 0.00937,-4.6610035 0.1395352,-0.024052 C 0.6388273,7.3314635 1.3550952,7.2775366 2.0419186,7.2579153 Z m 0.7798435,1.6195786 c -0.2821718,0 -0.5705236,0 -0.6449414,0.018988 -0.07132,0.015191 -0.1457431,0.036711 -0.1612477,0.045572 -0.018609,0.012659 -0.031027,0.5254704 -0.031027,1.3980474 v 1.379566 l 0.1178163,0.02532 c 0.2139543,0.04304 0.9364517,0.02405 1.1473045,-0.03038 C 4.0278379,11.50991 4.4402684,10.941526 4.3937571,10.142624 4.3596414,9.543478 4.0743403,9.156369 3.5192998,8.956612 3.3611612,8.8983811 3.2557341,8.8857222 2.8216253,8.8793928 Z m 8.2446083,0 c -0.284109,-0.012659 -0.574813,0 -0.725202,0.015191 -0.117829,0.015191 -0.235664,0.034179 -0.254269,0.04304 -0.031,0.012659 -0.04029,0.6514264 -0.04029,3.0571224 0,2.869771 0.003,3.04497 0.05578,3.066363 0.09613,0.04051 1.187606,0.02532 1.417063,-0.02152 1.240313,-0.242797 2.000003,-1.035497 2.232563,-2.328983 0.07132,-0.396476 0.09305,-1.013977 0.04965,-1.388933 C 13.74573,10.860256 13.665071,10.577583 13.485225,10.211868 13.342589,9.932359 13.286781,9.852482 13.041819,9.612722 12.669725,9.247134 12.322435,9.06282 11.751892,8.9336993 c -0.124032,-0.02785 -0.40153,-0.048104 -0.68564,-0.055699 z")
```
mondeja:
```suggestion
                path(d="M17.612 7.239H24l-.006.882-.01.885-2.006.012h-2.01v2.151h3.753v1.751h-3.752v3.841h-2.357V12Zm-7.653.02c.56-.024 1.078-.02 1.545.013.92.062 1.597.221 2.217.522.843.406 1.473 1.03 1.845 1.825.614 1.315.567 3.328-.109 4.661-.793 1.567-2.167 2.308-4.564 2.458-.713.046-2.226-.015-2.79-.11L8 16.61V7.435l.164-.022a22.974 22.974 0 0 1 1.795-.151zm-7.917 0c.687-.019 1.344 0 1.71.053.623.1 1.063.255 1.47.523.834.55 1.271 1.57 1.166 2.703-.078.802-.33 1.392-.8 1.872-.462.47-1.042.759-1.842.918-.301.059-.45.071-1.04.071h-.69l-.007 1.675L2 16.748l-1.002.013H0l.006-4.66.01-4.662.14-.024c.483-.084 1.2-.137 1.886-.157Zm.78 1.62c-.283 0-.57 0-.645.019a.968.968 0 0 0-.162.045c-.018.013-.03.526-.03 1.398v1.38l.117.025c.214.043.937.024 1.148-.03.778-.206 1.19-.774 1.144-1.573-.034-.6-.32-.987-.875-1.186-.158-.059-.263-.071-.697-.078Zm8.244 0a5.62 5.62 0 0 0-.725.015 1.973 1.973 0 0 0-.254.043c-.031.013-.04.651-.04 3.057 0 2.87.002 3.045.055 3.066.096.041 1.188.026 1.417-.021 1.24-.243 2-1.036 2.233-2.329a5.674 5.674 0 0 0 .05-1.389c-.056-.46-.137-.743-.317-1.11-.142-.279-.198-.359-.443-.598-.372-.366-.72-.55-1.29-.68a4.624 4.624 0 0 0-.686-.055z")
```

public/index.pug
```diff
@@ -78,27 +78,36 @@ html(lang="en-US")
...
...24 0.123254,0.560573 0.33278,0.919336 0.739511,1.289345 0.634748,0.577399 1.670063,0.846472 2.66224,0.695115 l 0.443707,-0.07286 V 13.760855 12.869528 H 21.319272 20.610575 V 12.140756 11.412 H 22.305288 24 v 2.208695 2.208681 l -0.197201,0.06169 c -1.700876,0.487694 -3.746858,0.526932 -5.096466,0.100885 -0.634748,-0.201808 -1.078454,-0.454069 -1.577624,-0.908139 -0.930552,-0.840873 -1.368095,-2.107776 -1.195544,-3.486826 0.258829,-2.135802 1.867268,-3.5428478 4.375443,-3.8231494 0.49917,-0.06165 1.596112,-0.033625 2.193883,0.056067 z M 9.2960508,8.0877595 c 0.264993,0.751184 1.4666972,4.6920635 1.6454122,5.4095915 0.110924,0.459699 0.221854,0.829691 0.240342,0.829691 0.01848,0 0.09861,-0.269088 0.178715,-0.60544 0.147898,-0.605426 0.893577,-3.015946 1.503673,-4.8602319 l 0.332781,-1.0090435 1.115429,-0.016828 1.11543,-0.011167 -1.534486,4.1482874 c -0.844276,2.287174 -1.552974,4.181928 -1.571462,4.209985 -0.02465,0.02802 -0.560796,0.05044 -1.232519,0.05044 H 9.8938228 l -0.277317,-0.77361 C 9.4686048,15.039001 8.8092048,13.177871 8.1621328,11.32797 7.5088987,9.4780539 6.9604261,7.9308344 6.9481007,7.8915957 6.9234524,7.8411439 7.1699555,7.8243309 8.057369,7.8243309 h 1.1400808 z")
```
mondeja:
```suggestion
                path(d="M4.495 7.746c.487.04 1.387.308 1.387.403 0 .107-.382 1.374-.425 1.424-.025.023-.222-.022-.45-.095-.487-.163-1.38-.28-1.756-.23-.703.084-1.097.415-1.048.863.05.432.314.606 1.645 1.105 1.763.65 2.404 1.317 2.416 2.494.006 1.228-.813 2.12-2.28 2.478-.58.14-1.892.168-2.6.05-.463-.072-1.276-.313-1.381-.397-.037-.04.296-1.475.363-1.553.007-.011.191.05.407.129.814.302 1.861.426 2.52.297.401-.079.771-.381.833-.684.123-.583-.272-.908-1.72-1.413-1.134-.398-1.904-.992-2.194-1.693C.01 10.437.052 9.73.311 9.22c.438-.835 1.405-1.379 2.7-1.508.412-.04.776-.033 1.484.034Zm18.007.084c.58.078 1.19.263 1.19.359 0 .05-.086.386-.197.762-.179.617-.21.673-.32.645-.075-.017-.352-.079-.623-.146-.604-.146-1.794-.174-2.311-.062-1.627.365-2.44 1.688-2.058 3.353.123.56.332.919.739 1.289.635.577 1.67.846 2.662.695l.444-.073V12.87h-1.417v-1.458H24v4.417l-.197.062c-1.701.488-3.747.527-5.097.101a3.912 3.912 0 0 1-1.577-.908c-.93-.841-1.368-2.108-1.196-3.487.259-2.136 1.867-3.543 4.376-3.823.499-.062 1.596-.034 2.193.056zm-13.206.258c.265.75 1.467 4.692 1.645 5.41.111.46.222.829.24.829.02 0 .1-.269.18-.605.147-.606.893-3.016 1.503-4.86l.333-1.01 1.115-.017 1.116-.01-1.535 4.148a344.932 344.932 0 0 1-1.571 4.21c-.025.028-.56.05-1.233.05H9.894l-.277-.774c-.148-.42-.808-2.281-1.455-4.131-.653-1.85-1.202-3.397-1.214-3.436-.025-.05.222-.068 1.11-.068h1.14z")
```

mondeja (changes requested):
These new icons should be optimized with [SVGO](https://jakearchibald.github.io/svgomg/).

public/index.pug
sachinraja: Also note that [`aria-labelledby` does not support two elements with the same id](https://web.dev/duplicate-id-aria/).

mondeja:
Please, don't request my review, I'm so busy for now.

public/stylesheet.css
ericcornelissen: @jorgeamadosoria, did you see my latest comment?

jorgeamadosoria: I hadn't.
Outline didn't appear, after fixing the menu in the other PR I saw what that is. I can restore outline, no problem

ericcornelissen (commented):
Please resolve the merge conflicts :slightly_smiling_face:

jorgeamadosoria:
The longer this PR stays unmerged the hardest it is to keep it updated.
Could we please get a definitive review? It's been months now.
@ericcornelissen @simple-icons/maintainers

sachinraja (approved):
This looks good to me and it works well, so if any other changes are necessary, they can be made in a separate PR.

PR #115, opened 2021-09-21
POC for pdf generation in the client side

jorgeamadosoria:
This is a POC for PDF generation on the fly in the client side.
The code is extremely sloppy, but the idea is to prove how it could be done.

It adds pdfkit.js, blob-stream.js and svg-to-pdfkit.js as standalone files from their releases in github (this is one of the ways they recommend to use it for browser use, the other one is through webpack).

The pdf.js code does an awful thing: it listens on the a.href to see if it is '#' (meaning it has never been clicked). If it is, it generates the pdf from the base64 svg string already existing in the html, then it changes the href and simulates a second click on it to actually download the pdf file. If the link is clicked again, the href is already modified and will download without  generating again. 
The problem with that (other than the obvious one with the roundabout way it's done) is that if the user generates too many blobs for download the page will keep using memory until their browser crash. I don't really think that's a problem, since downloading PDFs is already a fringe use case, not to mention download many pdfs one by one, but it is a posibility.

So, Pro / Con:
Pro: 
- The PDF repository, as well as all the necessary CI infra, github configurations, etc, would be unnecessary now. 
- Generating PNGs (a long term goal of mine) would be easier since it wouldn't need a separate repo like the pdf one, they could be generated on the fly.
- The pdfs are always in sync with the svg main repo without attention or automation.
- There is no additional use of server resources since the processing is on the client machine. This means that only the people that want a PDF will incur in the expense of getting it (which is not really a huge one, at least on my machine). 
- The pdf creation code is alongside the rest of the code for the website. This is a pro or a con depending on how it's looked at. It's a pro for me: all the code in an easily accesible repo is good for maintenance, specially in cases of personnel turnover.

Con:
- It adds weight to the website, since the pdfkit dependencies are now injected into the code that goes to the client. That's more obvious now, but I think this problem will persist even if we move those into webpack. I personally don't care, but others may.
- The code is not production ready just yet. We would need to iterate on this solution a lot before we get into a code that is not as sloppy as this one. I don't know enough webpack just yet, but I'm willing to try.
- There are no e2e tests or unit tests just yet. Those would be added if this goes forward.
- The aforementioned potential memory usage in the client if they download too many icons in pdf form. Not sure if this is me being paranoid or if it can happen, but it sounds as it could.
- This code does not help to offer a NPM package for the pdf files, something that was discussed in a separate issue that I don't recall right now. I don't think that's a huge loss, but others may.

I think this requires the input of as many @simple-icons/maintainers as possible, since it is a pretty big change (even if it is a simple one codewise).


adamrusted:
So my initial thoughts on this.
Cutting the manual work out of the PDF repo itself is a great idea. The less we have to do across multiple repos for a release, the better.

Adding the "bulk" to the website repo I'm not so much a fan of. The website is already slow enough (at least for me) without adding extra code that wouldn't be necessary if PDFs were imported as opposed to generated. I know the PDFkit files won't add much in the grand scheme of things, but I would like for us to focus first and foremost on getting the load time _down_ before we add anything else to the website code.

Again, this is void of context as the PDF repo was a thing before I joined, but I imagine there will be use cases for the PDFs without wanting to browse the website for them. Perhaps this is an argument to make the PDF repo its own NPM module, but I think all of our icons should be supplied as separate packages, with the website being our way of presenting them all - and not doing any of the heavy lifting itself.

jorgeamadosoria:
> Adding the "bulk" to the website repo I'm not so much a fan of. The website is already slow enough (at least for me) without adding extra code that wouldn't be necessary if PDFs were imported as opposed to generated. I know the PDFkit files won't add much in the grand scheme of things, but I would like for us to focus first and foremost on getting the load time _down_ before we add anything else to the website code.

Adding weight to the website is a problem indeed. However, the main driver of that is the front-loading of all icons into the generated HTML. Until that's dealt with, nothing else is really significant, I don't think. Besides, there are ways to reduce the weight added, perhaps with a custom version of the added files. After all, we are using an extremely small subset of features, we likely don't need more than half of the libraries' code.

> Again, this is void of context as the PDF repo was a thing before I joined, but I imagine there will be use cases for the PDFs without wanting to browse the website for them. Perhaps this is an argument to make the PDF repo its own NPM module, but I think all of our icons should be supplied as separate packages, with the website being our way of presenting them all - and not doing any of the heavy lifting itself.

I don't see a use case for a pdf icon at all, let alone for a NPM package of pdf icons. That's probably just a failure of imagination on my part, but still I don't put a lot of importance on pdf icons even existing. PNGs and SVGs are so much more useful for... well, everything, really. Even for building PDFs. However, just to keep the feature, an on the fly generation seems like a good compromise for me. Alternatively, creating the PDFs on webpack when releasing the website is another option, although I don't really think it's better weight-wise.

I suppose that I very much prefer to expand the website rather than to expand vertically to different repositories and technologies that need to be kept in sync and requires more knowledge to keep running than simply handing a website. Right now to keep the organization working we need to know how to work on icons, a website (without frameworks or libraries, I may add) github actions and workflows, node.js etc. That's in addition to knowing how all that is setup to keep it working when people leave or step back.

Keeping things as simple as possible is a great advantage in handling turnover and simplify onboarding of new people. And maintaining the project, too: it's not the same having to maintain 4 repositories (main, pdf, font and website) than maintaining only main and website (font is going to be dealt with in due time, not the subject of this PR).

ericcornelissen:
Did you consider generating the PDFs at build time (using the code of the simple-icons-pdf repo) and keeping the website as is?

jorgeamadosoria:
> Did you consider generating the PDFs at build time (using the code of the simple-icons-pdf repo) and keeping the website as is?

I did, briefly. It is certainly an option that can be explored too.

dirien:
Was there a request for PDF generation? I would never thought about this. SVG and PNGs for the win.

adamrusted:
I believe a request for PDF generation came in before I started (so over a year ago now).
It may be worth asking the wider community if it receives much use. We don't have NPM downloads to use as a metric, so would have to be a manual enquiry (somehow) - but we could look to deprecate the PDF offering entirely if there isn't enough of a need for it. We could then look at a PNG format, and whether that could be provided as a separate NPM module (I can see there being a need for this being a module as opposed to a web download) or as an extension of the main module, allowing people to choose their file format when requesting an icon, but storing the icons in SVG format.

jorgeamadosoria:
> I believe a request for PDF generation came in before I started (so over a year ago now).
> It may be worth asking the wider community if it receives much use. We don't have NPM downloads to use as a metric, so would have to be a manual enquiry (somehow) - but we could look to deprecate the PDF offering entirely if there isn't enough of a need for it. We could then look at a PNG format, and whether that could be provided as a separate NPM module (I can see there being a need for this being a module as opposed to a web download) or as an extension of the main module, allowing people to choose their file format when requesting an icon, but storing the icons in SVG format.

I like the idea of having a function in the main repo that converts svgs to pngs, so the internal storage is only svg and the pngs are served as a transformation on the base storage. This is something we can discuss further elsewhere, but I like it in principle. 

About PDFs, I would love to deprecate and remove that entirely, but since it was a request, I would hate to remove something that people are really using and we don't know about. How can we ask that? Should we ask that? 

As for @ericcornelissen's idea of generating the PDFs during build time, what do you think @adamrusted, @dirien ?

dirien:
@jorgeamadosoria, I am for build time too! I have too, to wait 1-2s before i can enter a search term into the input field, so aditional latency will be a bad UX.

sachinraja:
I'm for deprecating PDF generation after asking the community.

> About PDFs, I would love to deprecate and remove that entirely, but since it was a request, I would hate to remove something that people are really using and we don't know about. How can we ask that? Should we ask that?

We can tell them to fork it if they really need it.

adamrusted:
> About PDFs, I would love to deprecate and remove that entirely, but since it was a request, I would hate to remove something that people are really using and we don't know about. How can we ask that? Should we ask that? 

I guess we could build the PDFs up to a certain release of the main repo, and then anything new doesn't get a PDF generated (perhaps remove the PDF option on the website, if the file doesn't exist?). We can gauge use of the PDFs based on how many people ask for the PDF of a specific brand. If there is little / no feedback, we can trial a website release without PDFs all together, and see what feedback we get there.

> As for @ericcornelissen's idea of generating the PDFs during build time, what do you think @adamrusted, @dirien ?

Generating at build time could be the solution, assuming we would then not be loading the additional JS at load time?
This however may be out of scope if we decide to go with the above option.

jorgeamadosoria:
> Generating at build time could be the solution, assuming we would then not be loading the additional JS at load time?
> This however may be out of scope if we decide to go with the above option.

if we generate at build time then yes, the current added js won't be necessary, the weight of the website would be kept as is. It would increase the build time, I suppose, but that's not as important.

How about this: we provide SVGs, and in the main repo we put a script that generates the pdfs on demand,  something like `npm run export-pdf`.

Then we document that in the README, and deprecate the pdfs starting in the next release.
That way we don't increase the website size, we don't change the build process, we deprecate the pdf repo and we still serve pdfs to whoever really wants them (albeit in a little more involved way). What do you think?

adamrusted:
> How about this: we provide SVGs, and in the main repo we put a script that generates the pdfs on demand, something like `npm run export-pdf`.
> 
> Then we document that in the README, and deprecate the pdfs starting in the next release.

Sounds good to me, though obviously will need input from other @simple-icons/maintainers.

My proposed roadmap for the deprecation would be as follows:
- Remove PDFs and all other files from `simple-icons-pdf` repo. Replace with a README.md which explains PDFs have moved to the main repo, and while they won't be exported as a package (no different to currently), they are available via the command. As far as I'm concerned, this can happen as of the minor release which includes the script.
- Full deprecation of the PDF repo, and the command in the main repo, should probably be pushed to v7. That way we give any users of the PDF functionality a full release before removing the functionality all together (so early Summer 2022).

Just to make it end-user-proof, could whatever script you add save them to a /pdf directory, which gets added to the .gitignore, so people don't accidentally push thousands of PDFs with their PRs?

jorgeamadosoria:
> > How about this: we provide SVGs, and in the main repo we put a script that generates the pdfs on demand, something like `npm run export-pdf`.
> > Then we document that in the README, and deprecate the pdfs starting in the next release.
> 
> Sounds good to me, though obviously will need input from other @simple-icons/maintainers.
> 
> My proposed roadmap for the deprecation would be as follows:
> 
> * Remove PDFs and all other files from `simple-icons-pdf` repo. Replace with a README.md which explains PDFs have moved to the main repo, and while they won't be exported as a package (no different to currently), they are available via the command. As far as I'm concerned, this can happen as of the minor release which includes the script.
> * Full deprecation of the PDF repo, and the command in the main repo, should probably be pushed to v7. That way we give any users of the PDF functionality a full release before removing the functionality all together (so early Summer 2022).
> 
> Just to make it end-user-proof, could whatever script you add save them to a /pdf directory, which gets added to the .gitignore, so people don't accidentally push thousands of PDFs with their PRs?

Ok, I like it.
I would create a /pdf folder at the same level as /icons. /pdf in gitignore, yes. the command of course won't be added to any workflow, since it is entirely optional.
And since the dependencies will be added to the main repo, the website won't increase its weight.

Now, we have a decision to make regarding pdf downloading from the website: completely out of the HTML offering, I suppose?

This is of course all pending on @simple-icons/maintainers input.

ericcornelissen:
The reason the PDF thing exists in the first place is basically just because of https://github.com/simple-icons/simple-icons/issues/743. I don't recall anything being said about PDFs by the community other than that issue.

---

If I'm reading this right, it seems the current proposal is to add a script to [the simple-icons/simple-icons repo](https://github.com/simple-icons/simple-icons) that generates the PDFs on demand. But then what? That's not useful to anyone except people that clone that repository. An npm script like that isn't really runnable by anyone installing [the simple-icons npm package](https://www.npmjs.com/package/simple-icons) as far as I know. And adding an executable `.js` file to generate PDFs on-demand to  [the simple-icons npm package](https://www.npmjs.com/package/simple-icons) seems like a bad idea to me. As it stands we think PDFs provide very little value, but this solution would introduce a ton of dependencies and a binary to maintain to the main project.

jorgeamadosoria:
> If I'm reading this right, it seems the current proposal is to add a script to [the simple-icons/simple-icons repo](https://github.com/simple-icons/simple-icons) that generates the PDFs on demand. But then what? That's not useful to anyone except people that clone that repository. 

Yes, correct. A kind of advanced user, if you will. This builds on @sachinraja's idea to suggest forking if someone needs PDFs (but makes it easier), and if someone needs PDFs for the use case outlined in #743, then they have an easy way to do it. 
It's a win - win: We don't have to support anything other than a pdf script (with code already written and known dependencies) and the user that needs the pdfs can get them.  Not a huge problem.

sachinraja:
I don't think that is worth the trouble. If a user wants it, they can just fork the PDF repo like I said above.

jorgeamadosoria:
The options weighed so far are, in no particular order:

1 - Keep the repo as it is right now.
2 - Generate the PDFs during build time from the website.
3 - Generate the PDFs on the fly in the website (the code in this PR).
4 - Generate the PDFs in a command line from the main repo.
5 - Deprecate the PDF repo altogether without replacement.

I prefer options 3 or 4. 
@sachinraja seems to prefer 5.

@ericcornelissen @adamrusted @dirien ?

ericcornelissen:
An ideal solution in my eyes would be 1. 2 is probably the easiest solution if we want PDFs. I'm not a fan of 3. 4 is a bad idea in my opinion. 5 is okay to me, but may end up being just be a temporary solution if there's demand for PDFs in the future.

sachinraja:
> may end up being just be a temporary solution if there's demand for PDFs in the future

If there's demand for PDFs, someone can fork the repo and maintain it themselves. There are many simple-icons packages doing the same thing for React components, Vue components, etc.

Just going off [the traffic for simple-icons-pdf](https://github.com/simple-icons/simple-icons-pdf/graphs/traffic) there doesn't seem to be many people looking at the repo. There may be a different story for downloads from the website though.

jorgeamadosoria:
> > may end up being just be a temporary solution if there's demand for PDFs in the future
> 
> If there's demand for PDFs, someone can fork the repo and maintain it themselves. There are many simple-icons packages doing the same thing for React components, Vue components, etc.
> 
> Just going off [the traffic for simple-icons-pdf](https://github.com/simple-icons/simple-icons-pdf/graphs/traffic) there doesn't seem to be many people looking at the repo. There may be a different story for downloads from the website though.

sure, and they will always be able to do it. But wouldn't it be better to just provide the npm command so if they just want the pdfs at a moment on time? Specially if we already have the code written...

In any case, we know where we all here stand.

We need the other @simple-icons/maintainers to chime in. I think this is big enough to require input from everyone or at least most of us.

ericcornelissen:
I definitely agree we don't _have_ to support PDFs, but if we would want to in the future removing them now is just a temporary fix. Not to mention that PNGs pose the exact same challenges and from past discussion it seems more people are on board with providing support for those.

> But wouldn't it be better to just provide the npm command so if they just want the pdfs at a moment on time?

I'm off the opinion that, especially in the npm ecosystem, providing an npm command which requires users to download/clone the git repository first is a bad idea. It's not a normal workflow and not an ideal workflow for automation either, if you want to provide _ something_ the PDF repository is already there and can be used as such. If we decide to remove PDFs from the website, the current build-chain (in terms of open PRs) is sufficient to provide PDFs completely automatically.


> Specially if we already have the code written

This is why archiving the repo may be a better idea than removing the repo. The code will still be available for anyone to look at.

jorgeamadosoria:
> I definitely agree we don't _have_ to support PDFs, but if we would want to in the future removing them now is just a temporary fix. Not to mention that PNGs pose the exact same challenges and from past discussion it seems more people are on board with providing support for those.

The current repository requires a level of maintenance and introduces a level of complexity to the project that, specially with the announced departures, it's not something that can be supported. It is unnecessarily complex and unnecessarily separated just for the sake of separation. 
This all stems from the concept of having a website that is only about showcasing the things that exist and void of dynamic functionality. This is, again, a restriction that forces the project to do this kind of backend gymnastics.
A website free from the self imposed limitations of not using libraries, having to support non JS users, and weight watching continuously could take care of this and many other things it needs (like UI aesthetic improvements and the very necessary pagination) very quickly and far easier than now.

PNGs will go the way PDFs goes, that's why there is no need to discuss them now.

> 
> > But wouldn't it be better to just provide the npm command so if they just want the pdfs at a moment on time?
> 
> I'm off the opinion that, especially in the npm ecosystem, providing an npm command which requires users to download/clone the git repository first is a bad idea. It's not a normal workflow and not an ideal workflow for automation either, if you want to provide _ something_ the PDF repository is already there and can be used as such. If we decide to remove PDFs from the website, the current build-chain (in terms of open PRs) is sufficient to provide PDFs completely automatically.

The command is not for people that just want to download a PDF. It's for users that know how to fork, clone, etc., but don't really want to. It's easier to clone the main repo and run a command than to fork. It's a convenience for people, and it assumes that the PDF repository is not going to be there. The idea is to support as few artifacts and codebases as possible.

-------------------

Since this is a conceptual discussion rather than a technical one, I don't think there is space for change for the time being. Without further input from the @simple-icons/maintainers that haven't weighed in, this issue won't move forward.

ericcornelissen:
If other maintainers don't weigh in, you just make a decision. It can always be changed later and all the work already done is here on GitHub for your future self to use.

I won't go into the rest of your comment, if you disagree with the architecture of the website than you have all the power you need to change it.

jorgeamadosoria:
> If other maintainers don't weigh in, you just make a decision. It can always be changed later and all the work already done is here on GitHub for your future self to use.
> 
> I won't go into the rest of your comment, if you disagree with the architecture of the website than you have all the power you need to change it.

I know I can. But I won't if I'm. the only one that wants to do it. If no other @simple-icons/maintainers intervenes in say, the next month, I will close the ticket, forget all about the website and continue working on my own alternative. That'll be easier and will give me more time to focus on the main repo, which is where most of the pending content is.

sachinraja:
Curious as to what you have in mind. What would you like the website to be built on?

jorgeamadosoria:
> Curious as to what you have in mind. What would you like the website to be built on?

I was thinking on a couple options:
- adding jquery or something similar to handle all the JS more expeditiously but keeping the rest of the website fundamentally the same,
- Refactor it all with React, possibly adding a publishing step with Gatsby. 

But more important than technology is changing the approach of the website from showcasing icons in a completely static manner to a place where users can interact with the icons (thus allowing for more dynamic features to be added, like the PDF generation, ordering with JS and not CSS, having several pages including the icon detail view, etc.) The concept is more akin to the Font Awesome icons page than what exists now.

sachinraja:
I've been thinking about this too and I actually even made a basic site with Gatsby. I'm up for that because it does allow for more interactivity and other features. jQuery wouldn't really help, it's not really useful for modern JS.

I think we should continue this discussion somewhere else, perhaps in another issue.

service-paradis:
There are a lot of things going on here...

When adding the new design, there was a header that asks people opinion on it. Maybe we could do something similar for PDF deprecation. We could create a Github discussion about it and tell people that visits the website to give their opinion. This way, we could "measure" if this feature is useful or not and talk about current use case in the discussion.

If there are some justified use case, we could plan on a solution to replace the current repo. If there are none, we could just archive it and remove the feature from the website.

Currently, it is great to see this POC, but I'm not yet convinced this is the ideal solution. The plan is to remove complexity. This solution might remove a complex repo, but it adds complexity to a simple one. The flow is quite complex and I'm unsure of the different browsers and devices support (or behavior).

If we really need and want to support PDF, I think I would be more in favor of generating them all at build time (maybe in the main repo). The website would still be able to serve them. They could even be available with `unpkg` for example (`https://unpkg.com/simple-icons@v5/pdf/[ICON SLUG].pdf`)

ghost (commented):
It looks good with me.

jorgeamadosoria:
> Currently, it is great to see this POC, but I'm not yet convinced this is the ideal solution. The plan is to remove complexity. This solution might remove a complex repo, but it adds complexity to a simple one. The flow is quite complex and I'm unsure of the different browsers and devices support (or behavior).

It adds complexity to the website, yes. Hence my concern with the conceptual tenets of the website as a showcase rather than an interactive website. I don't see adding complexity to the website as something inherently bad when it's not avoidable. Keeping one repo simple at the expense of supporting others is not reducing the overall complexity of things, which is the goal.

The flow is a POC. It can stand a lot of refactoring. As to different browser support, I'm pretty sure it is well supported (yet another benefit of using libraries that I had overlooked) but it's just a matter of testing it and fixing it appropriately. I wouldn't hold neither of these things against the idea of generating dynamic pdfs, not just yet. 
Maybe after improving the code a little bit, if it is still complex or brittle. But certainly not just yet, in its pre-alpha form.


> If we really need and want to support PDF, I think I would be more in favor of generating them all at build time (maybe in the main repo). The website would still be able to serve them. They could even be available with `unpkg` for example (`https://unpkg.com/simple-icons@v5/pdf/[ICON SLUG].pdf`)

Generating PDFs in the main repo is obviously the simpler choice. And if that's agreed by majority I would be ecstatic to remove this and rework the website to download them (and the upcoming PNGs) directly from the main repo.
Unfortunately this was discussed in some other issue, I'm sure @ericcornelissen remembers best. It was rejected on the grounds of it adding size to the npm package or something like that, I don't fully remember. I think it was the issue that created the PDF repo.

I strongly favor the static generating of everything in the main repo. That's the simplest, most direct option. But the fact that when requested a new repo was created tells me that not everyone feels that way, hence we would need to make a majority decision, for which, again, we need more @simple-icons/maintainers here, if not all of them, to vote on this conclusively and take a position.

ericcornelissen:
Roughly @jorgeamadosoria, but that's just my opinion - as I've said before, if there's consensus against that then you can do that.

Anyway, here's my current thoughts on the various options on the table for PDFs that I see.

- Generating PDFs on the client side: again, I see no reason for this. The content of the PDF is not dynamic; it might break due to the client environment; it increases the amount of JavaScript client side; possibly more(?). We can generate the PDFs before or during the website build. This is safer, less likely to break, and easier to debug.
- Generating the PDFs in the website repository at build time: one of the simplest solution. The only drawback really is that the collection of PDFs is not easily reusable by others.
- Generate the PDFs in its own repo (like [the font](https://github.com/simple-icons/simple-icons-font)): this is really not that hard and the font proves that it's a sustainable approach. The only problem we faced here is that triggering the CI of the various repos is somewhat difficult (which is the only disadvantage of not having a monorepo we have, I should add).

  Actually, in this regard,, I liked @service-paradis' [comment](https://github.com/simple-icons/simple-icons-website/pull/115#issuecomment-927849831). If we stick with the separate PDF repo and the CI workflow already ready in PRs, publish the PDFs to npm, add unpkg or jsdelivr URLs to the website, then we have a solution. (I have indeed said before that publishing the PDFs to npm seemed unnecessary, but that was under the assumption no-one would use them - if we use them for the website this way than there's a use).

   To clarify, the problem we faced with the build is this:
   1. The website needs both the SVGs and the PDFs
   2. When new SVGs are released, a build for the font and PDFs can be triggered to release an update in accordance with the new SVGs
   3. Now, at what point is a rebuild for the website trigger? If the SVGs workflow does this, the repo's may not have been updated. We could move it to the PDF repo, but this is only a temporary solution (e.g. what if a repo for PNGs is created?).

- _Generate the PDFs in [the main repo](https://github.com/simple-icons/simple-icons)_ ...
  -  But don't publish them: this seems pointless to me. Using git for packaging is simply not ideal. If it's about having the script available to others, just put it in it's own repo and publish that.
  - _And publish it_ ...
    - With the SVGs: this seems like a waste to me. People that need the SVGs probably don't need the PDFs and vice versa. It's arguably not a huge problem (it's not gigabytes of content, after all), but we might as well avoid it in my opinion - even if the alternative is a bit more work from our side.
    - As its own package: definitely possible, would solve the CI problem. However, it's not in line with the organisation setup. Is that a problem? That depends on your viewpoint. It would be possible to combine all the repos together into a monorepo.
- Remove the PDFs: easy solution and probably fine with our users. However, as I mentioned before, this problem/discussion may come back in the future (e.g. for PNGs)

And my list of favorites

1. Separate repo for PDFs, publish to npm, unpgk/jsdelivr links in the website
2. Generate PDFs at build time
3. No PDFs
4. Monorepo

jorgeamadosoria:
This discussion has become pointless. Let's keep whatever exists now, and I'll retake this discussion in a few months if necessary.

service-paradis:
We're a group of maintainers and every member might have different opinions on different subjects. Expressing ourselves and discussing is not pointless and every opinions should always be welcome.

sachinraja:
I don't see how waiting a few months is helpful. Can we reopen this or at least create a new issue?

jorgeamadosoria:
> I don't see how waiting a few months is helpful. Can we reopen this or at least create a new issue?

feel free to open a new ticket or take ownership of this one. 
I have exhausted my arguments. If I can't gather majority support for any of the options, then it stands to reason that the status quo must remain until a significant number of maintainers can agree on one of the options. 
So the PDF repo will stay, broken as it may be, unless someone take steps to fix it so it at least becomes functional again.

PR #123, opened 2021-10-17
Button design

### LINKED ISSUE #46: [Discussion] Update controls design
### Kind of issue 

- [ ] Feature
- [ ] Bug
- [x] Design
- [ ] Other, namely:

### Description

These are suggestions for changing the design of the controls.

- Move color mode buttons away from the other controls.
- Omit borders, e.g.:
  preview no brders
- As per [this comment](https://github.com/simple-icons/simple-icons/discussions/4865#discussioncomment-365975), grouped buttons, e.g.:
  preview grouped buttons

jorgeamadosoria:
Complying with @fbernhart comment, this PR removes the margin between the buttons but keep a thin border of separation.

- Removed the internal borders for button groups.
- Rounded only the first and last buttons in the relevant corners (not internal)
- To simplify the CSS, I had to move sorting relevance to the middle of the sort group, because it is sometimes hidden and the last-of-type selector does not work in that case.


public/stylesheet.css
```diff
@@ -397,13 +396,23 @@ body.no-js .control-field__title {
 }
 
 .control__button {
+  border-bottom: 0.1rem solid var(--color-button-border);
+  border-top: 0.1rem solid var(--color-button-border);
+  border-right: 0.1rem solid var(--color-button-border);
```
ericcornelissen: This can be simplified to avoid repetition

```suggestion
  border: 0.1rem solid var(--color-button-border);
  border-left-width: 0;
```

With this change you can also have just `border-left: 0.1rem` in the `:first-of-type` ruleset.

public/stylesheet.css
```diff
@@ -374,14 +374,13 @@ body.no-js .control-field__title {
...
   background-color: transparent;
-  border: 0.1rem solid var(--color-button-border);
-  border-radius: var(--border-radius-small);
   color: var(--color-text-default);
   height: 2.4rem;
   padding: 0.4rem;
 }
 
 .control__input {
+  border: 0.1rem solid var(--color-button-border);
```
ericcornelissen, line 383: Did you intentionally leave out the `border-radius` here?

public/index.pug
```diff
@@ -81,12 +81,12 @@ html(lang="en-US")
...
-              svg(role="img" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" aria-labelledby="order-color")
-                path(d="M9.219 18.857H6.648V.429A.414.414 0 0 0 6.219 0H3.648a.418.418 0 0 0-.308.121.418.418 0 0 0-.121.308v18.428H.648a.4.4 0 0 0-.402.268c-.071.17-.04.326.094.469l4.286 4.287c.098.08.2.119.308.119a.447.447 0 0 0 .308-.119l4.272-4.273a.506.506 0 0 0 .134-.32.417.417 0 0 0-.429-.431zm3-15.428h3.429A.412.412 0 0 0 16.076 3V.429A.416.416 0 0 0 15.648 0h-3.429a.414.414 0 0 0-.429.429V3a.414.414 0 0 0 .429.429zm0 6.857h6a.42.42 0 0 0 .309-.12.42.42 0 0 0 .121-.308V7.286a.418.418 0 0 0-.121-.308.417.417 0 0 0-.309-.121h-6a.414.414 0 0 0-.308.121.417.417 0 0 0-.12.308v2.572c0 .125.04.228.12.308a.42.42 0 0 0 .308.12zm0 6.857h8.572c.125 0 .229-.039.309-.119s.119-.184.119-.309v-2.572c0-.125-.039-.227-.119-.307s-.184-.121-.309-.121h-8.572a.418.418 0 0 0-.308.121.415.415 0 0 0-.12.307v2.572c0 .125.04.229.12.309.081.08.183.119.308.119zm11.451 3.55a.408.408 0 0 0-.307-.121H12.219a.416.416 0 0 0-.429.428v2.572c0 .125.04.227.121.309a.42.42 0 0 0 .308.119h11.144a.414.414 0 0 0 .307-.119.424.424 0 0 0 .121-.309V21a.416.416 0 0 0-.121-.307z")
```
ericcornelissen: I'm not convinced this is the solution we should go with, the way the button order changes when searching as a result of this is confusing in my opinion.

We could insert the button for sorting by relevance only when there's an active query. Alternatively we could use classes here to overwrite (or replace) the use of `:last-xxx`. What do you think @jorgeamadosoria?

public/stylesheet.css
```diff
@@ -397,13 +396,23 @@ body.no-js .control-field__title {
...
-  margin: 0 0.3rem;
   width: 2.4rem;
 }
 .control__button:first-of-type {
   margin-left: 0;
+  border-left: 0.1rem solid var(--color-button-border);
+  border-top-left-radius: var(--border-radius-small);
+  border-bottom-left-radius: var(--border-radius-small);
+}
+.control__button:last-of-type:not(.hidden) {
```
ericcornelissen: The last part of this selector doesn't do anything as far as I can tell.

```suggestion
.control__button:last-of-type {
```

public/stylesheet.css
```diff
@@ -397,13 +396,23 @@ body.no-js .control-field__title {
...
+  margin-right: 0;
```
ericcornelissen: I don't think this is necessary, nor the `margin-left` in the `:first-of-type` ruleset.

public/stylesheet.css
jorgeamadosoria: no. Good catch

public/index.pug
jorgeamadosoria: I don't think this is a big issue, but I see your point.
I don't know if inserting the button on an active query is going to work, but maybe a CSS class will. I'll try that approach

public/stylesheet.css
jorgeamadosoria: that is true. I first tried to use this to exclude the hidden button before changing its position. We can remove it.

public/stylesheet.css
jorgeamadosoria: this may go away if we build a dedicated CSS class anyways.

public/stylesheet.css
jorgeamadosoria: ok

public/stylesheet.css
jorgeamadosoria: restored

public/index.pug
jorgeamadosoria: reworked by applying a class in search.js whenever the search string changes.

jorgeamadosoria:
ok, all suggestions worked into this PR.

public/stylesheet.css
ericcornelissen: Looks like it's still here 🤔

public/stylesheet.css
```diff
@@ -397,13 +397,23 @@ body.no-js .control-field__title {
...
-  margin: 0 0.3rem;
   width: 2.4rem;
 }
-.control__button:first-of-type {
+.first__button {
   margin-left: 0;
+  //border: 0.1rem solid var(--color-button-border);
+  border-left: 0.1rem solid var(--color-button-border);
+  border-top-left-radius: var(--border-radius-small);
+  border-bottom-left-radius: var(--border-radius-small);
```
ericcornelissen: This looks off 🤔

jorgeamadosoria: make a suggestion? other than removing the comment, I don't know what looks off there.

jorgeamadosoria: I reworked this to reduce the class attribute. I changed the border-left for border-left-width, that way all the other border properties are inherited from control__button.
Maybe that looks better? It's certainly less redundant and shorter.

public/stylesheet.css
jorgeamadosoria: it's not anymore

public/scripts/search.js
```diff
@@ -84,12 +85,16 @@ export default function initSearch(history, document, ordering, domUtils) {
     if (query !== '') {
       domUtils.showElement($searchClear);
       domUtils.showElement($orderByRelevance);
+      domUtils.toggleClass($orderByRelevance, 'last__button');
+      domUtils.toggleClass($orderByColor, 'last__button');
```
ericcornelissen: Toggle here doesn't work with the current implementation because this if-branch is executed for every keystroke. As a result, after typing one character it's good, after typing another it's wrong, etc. (ignoring the de-bouncing for simplicity)

ericcornelissen (changes requested):

public/scripts/dom-utils.js
```diff
@@ -11,3 +11,21 @@ export function showElement($el) {
...
+  if ($el) {
+    $el.classList.remove(clazz);
+  }
+}
+
+export function toggleClass($el, clazz) {
+  if ($el) {
+    $el.classList.toggle(clazz);
+  }
+}
```
ericcornelissen, line 31: Should this be removed, it seems to not be used anymore?

tests/dom-utils.test.js
```diff
@@ -42,3 +47,25 @@ describe('::showElement', () => {
...
+describe('::toggleClass', () => {
+  beforeEach(() => {
+    $el.classList.toggle.mockReset();
+  });
+
+  it('toggles a CSS class for addition/removal', () => {
+    toggleClass($el, 'last__button');
+    expect($el.classList.toggle).toHaveBeenCalledWith('last__button');
+  });
+});
```
ericcornelissen: This is the exact same test twice.

Also, the add and remove class helpers should probably be tested as well/instead.

ericcornelissen (commented):
Functionality works as expected now, nice work :+1: Got two things that can be improved in the code

public/stylesheet.css
ericcornelissen: The main reason I commented here is because `// comment` is not a comment in CSS.

jorgeamadosoria: It's fine, it doesn't matter anymore

public/scripts/dom-utils.js
jorgeamadosoria: It's not used anymore, but I have plabs for it elsewhere. And this being an utils class, I thought that it doesn't hurt to add it now since it's already there

tests/dom-utils.test.js
jorgeamadosoria: Add and remove should be added, yes.
I have issues resolving " and ' with prettier, and that creates annoying merge conflicts where both branches are the same. Hence why you would find this kond of duplications.

public/scripts/dom-utils.js
ericcornelissen: Feel free to keep it then, your call

jorgeamadosoria: I will for now, thanks. It will help later on in some refacts I plan on doing soon when I'm done with the compact design.

tests/dom-utils.test.js
jorgeamadosoria: Done

tests/dom-utils.test.js
```diff
@@ -15,13 +24,9 @@ describe('::hideElement', () => {
...
   });
 
-  it('adds the .hidden class', () => {
+  it('adds the .hidden class and the aria-hidden attribute', () => {
     hideElement($el);
     expect($el.classList.add).toHaveBeenCalledWith('hidden');
-  });
-
-  it('adds the aria-hidden attribute', () => {
-    hideElement($el);
```
ericcornelissen, line 24: I don't understand why you made this change. What's the point of combining the two test case? Plus, if there's just one test case, why even have the `describe` and `beforeEach`?

My motivation for originally writing it as two separate tests is because it's testing two separate things. I don't care if you want to change the approach, but this Pull Request seems like an odd place to do that.

tests/dom-utils.test.js
```diff
@@ -32,13 +37,31 @@ describe('::showElement', () => {
...
+  it('removes the .hidden class and the aria-hidden attribute', () => {
     showElement($el);
     expect($el.classList.remove).toHaveBeenCalledWith('hidden');
+    expect($el.removeAttribute).toHaveBeenCalledWith('aria-hidden');
   });
+});
 
-  it('removes the aria-hidden attribute', () => {
-    showElement($el);
-    expect($el.removeAttribute).toHaveBeenCalledWith('aria-hidden');
```
ericcornelissen, line 42: I don't understand why you made this change. What's the point of combining the two test case? Plus, if there's just one test case, why even have the `describe` and `beforeEach`?

My motivation for originally writing it as two separate tests is because it's testing two separate things. I don't care if you want to change the approach, but this Pull Request seems like an odd place to do that.

ericcornelissen (approved):
Feel free to merge this as you see fit @jorgeamadosoria

tests/dom-utils.test.js
jorgeamadosoria: the two things are not really separate things. They can't be called by themselves: As a matter of fact, if they were to be called separate from each other this test would throw an exception, as it should since that shouldn't happen.We are testing the functionality of hideElement, which should have the two success criteria happening together. 
Agreed in that this PR is not the best place for this, but I was adding the other tests and I saw a chance to refactor this here.

tests/dom-utils.test.js
jorgeamadosoria: same comment as the previous conversation.

PR #124, opened 2021-10-18
Fast-fuzzy package for searching

### LINKED ISSUE #36: Improve website search
### Kind of issue 
  - [x] Improvement
  - [ ] Bug
  - [ ] Other, namely: 

### This issue concerns 
  - [ ] The NPM Package
  - [x] The website
  - [ ] Other, namely: 

### Description
Currently the website search performs poorly on some strings, not weighing exact matches high enough. For example, when you want to find Facebook by searching for "fa" it is preceded by the less accurate match "Fnac" (and several others):

image

I have implemented a [somewhat better search algorithm here:](https://codepen.io/birjolaxew/pen/moEmXO)

<p align="center"></p>

This search algorithm is based on the one described in [this blog post](https://www.forrestthewoods.com/blog/reverse_engineering_sublime_texts_fuzzy_match/), without some of the more fancy stuff. The bonuses/penalties for matching letters in specific positions have been tweaked based on our needs, putting more value on exact matches and on matches from the start of the string.

The big downside is that there is no easy way to filter out poor matches, because the score of each match depends on the length of the pattern and the length of the searched string. It could potentially be tweaked to better do this, but I don't have the time to do that right now.

Ideally we would implement this, or another search algorithm, on the website to give better search results.

jorgeamadosoria:
Added the https://www.npmjs.com/package/fast-fuzzy to the website for the website. While the code weight increases, the need for customized algorithms to fix the current fuzzy search bug disappears.


package.json
```diff
@@ -31,6 +31,8 @@
     "test:unit": "cross-env TEST_ENV=unit jest"
   },
   "dependencies": {
+    "fast-fuzzy": "1.10.10",
+    "graphemesplit": "2.4.4",
```
ericcornelissen: Why was this dependency added?

jorgeamadosoria: it's a dependency of fast-fuzzy

sachinraja: We don't have to add that manually. `npm` will manage it and resolve the dependency. You can safely remove it and you'll see that npm still resolves it in `package-lock.json`.

jorgeamadosoria: I don't think I added it manually. I just npm i fast-fuzzy.

jorgeamadosoria:
I've removed graphemesplit from the dependencies

package.json
mondeja: package.json dependencies objects are only populated when you explicitly install them:

```bash
$ npm init -y > /dev/null
$ npm i fast-fuzzy
$ cat package.json
{
  "name": "foo",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "fast-fuzzy": "^1.10.10"
  }
}
```

public/scripts/search.js
```diff
@@ -1,5 +1,6 @@
 import { ORDER_BY_RELEVANCE } from './ordering.js';
 import { decodeURIComponent, debounce, normalizeSearchTerm } from './utils.js';
+const { Searcher } = require('fast-fuzzy');
```
sachinraja: Can we use the ESM import style here? See the imports above this line

public/scripts/search.js
```diff
@@ -42,6 +31,8 @@ function setSearchQueryInURL(history, path, query) {
   }
 }
 
+const keySelector = (icon) => icon.querySelector('h2').textContent;
```
sachinraja: I think this should be removed

public/scripts/search.js
```diff
@@ -96,16 +90,16 @@ export default function initSearch(history, document, ordering, domUtils) {
...
       }
     }
-
+    const result = searcher.search(query);
     let noResults = true;
     $icons.forEach(($icon) => {
-      const brandName = $icon.getAttribute('data-brand');
-      const score = getScore(query, brandName);
-      if (score < 0) {
+      const brandName = $icon.querySelector('h2').textContent;
```
sachinraja: Can we use `$icon.getAttribute('data-brand')'` here like we did before?

public/scripts/search.js
```diff
@@ -13,21 +14,9 @@ function getQueryFromParameter(location, parameter) {
...
-    }
-
-    score += index;
-    index++;
-  }
-
-  return score;
+function getScore(results, iconName) {
+  const result = results.find((r) => r.original === iconName);
+  return result ? result.score : -1;
```
sachinraja: I think this should be removed

tests/e2e.test.js
```diff
@@ -255,7 +255,7 @@ describe('Search', () => {
     await $searchInput.press('Backspace');
 
     const $gridItemIfEmpty = await page.$('.grid-item--if-empty');
-    expect(await isHidden($gridItemIfEmpty)).toBeTruthy();
+    expect(await isHidden($gridItemIfEmpty)).toBeFalsy();
```
sachinraja: Why did this need to be changed?

sachinraja (changes requested):

public/scripts/search.js
jorgeamadosoria: It didn't work for me. I can try again.

tests/e2e.test.js
jorgeamadosoria: It didn't need to. That's a debug change that I forgot to rollback, and was hiding an error.
Good catch!

public/scripts/search.js
jorgeamadosoria: yes

public/scripts/search.js
jorgeamadosoria: yes

public/scripts/search.js
jorgeamadosoria: yes

sachinraja:
LGTM, but needs conflict resolution

sachinraja (approved):

jorgeamadosoria:
Merged in #153

PR #126, opened 2021-10-19
Remove JavaScript blurs

ericcornelissen:
Remove the use of `.blur()` to improve a11y, in particular when it comes to resetting the tabindex when clicking the control buttons.

### DIFF
--- a/public/scripts/color-scheme.js
+++ b/public/scripts/color-scheme.js
@@ -29,17 +29,14 @@ export default function initColorScheme(document, storage) {
   $colorSchemeDark.addEventListener('click', (event) => {
     event.preventDefault();
     selectColorScheme(COLOR_SCHEME_DARK);
-    $colorSchemeDark.blur();
   });
   $colorSchemeLight.addEventListener('click', (event) => {
     event.preventDefault();
     selectColorScheme(COLOR_SCHEME_LIGHT);
-    $colorSchemeLight.blur();
   });
   $colorSchemeSystem.addEventListener('click', (event) => {
     event.preventDefault();
     selectColorScheme(COLOR_SCHEME_SYSTEM);
-    $colorSchemeSystem.blur();
   });
 
   function selectColorScheme(selected) {
--- a/public/scripts/copy.js
+++ b/public/scripts/copy.js
@@ -16,7 +16,6 @@ export default function initCopyButtons(window, document, navigator) {
       event.preventDefault();
 
       const value = $colorButton.innerHTML;
-      $colorButton.blur();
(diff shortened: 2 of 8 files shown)


jorgeamadosoria:
> Remove the use of `.blur()` to improve a11y, in particular when it comes to resetting the tabindex when clicking the control buttons.

explain this one for my benefit, please? I don't know what blur has to do with accessibility.

ericcornelissen:
When you call `.blur()` on a node it removes the focus from that element and resets it to the top of the page - so if you do that upon clicking on the element it fires the event and then resets the focus to the top of the page. The reason this is bad for a11y is because it is annoying for keyboard users who go through the page content primarily by tabbing through focusable elements.

So, currently on the website, if you tab through to say the buttons for the dark/light mode and "press" (using enter or spacebar) one of them than the mode is changed but the focus is back at the top of the page. There's really no reason for this and it's much nicer if the focus stays where it is so that the keyboard user doesn't have to re-orient themselves as well as tab through a bunch of elements they already saw.

jorgeamadosoria:
> When you call `.blur()` on a node it removes the focus from that element and resets it to the top of the page - so if you do that upon clicking on the element it fires the event and then resets the focus to the top of the page. The reason this is bad for a11y is because it is annoying for keyboard users who go through the page content primarily by tabbing through focusable elements.
> 
> So, currently on the website, if you tab through to say the buttons for the dark/light mode and "press" (using enter or spacebar) one of them than the mode is changed but the focus is back at the top of the page. There's really no reason for this and it's much nicer if the focus stays where it is so that the keyboard user doesn't have to re-orient themselves as well as tab through a bunch of elements they already saw.

got it.
So the idea is to keep the focus in the just clicked element to make people able to tab along and not having to restart their navigation.
I have no problem with that.

jorgeamadosoria:
your code looks fine and to the point. IMO you can merge.

ericcornelissen:
> So the idea is to keep the focus in the just clicked element to make people able to tab along and not having to restart their navigation.

Correct


> your code looks fine and to the point. IMO you can merge.

Did you give a try locally, nothing buggy happens when you interact with the page on your end?

jorgeamadosoria:
yes, I tested it locally, I don't see a problem. 
The tabbing order (except the known issue when ordering by color) is fine.
I searched, chose the different orderings, and color schemes, downloaded a file, etc. At least it's not buggy, that's certain

ericcornelissen:
Thanks @jorgeamadosoria, merged it in now :ok_hand:

PR #127, opened 2021-10-23
Popup with Third party extensions

### LINKED ISSUE #16: Add (page for) first- and third-party packages/extensions

---

### Kind of issue 

- [x] Feature
- [ ] Bug
- [ ] Design
- [ ] Other, namely:

### Description

Add something to the website (a page, modal, other suggestions?) that lists the first-party packages/extensions ([npm package](https://www.npmjs.com/package/simple-icons), [packagist package](https://packagist.org/packages/simple-icons/simple-icons), [font](https://www.npmjs.com/package/simple-icons-font)) and [the (known) third-party packages/extensions](https://github.com/simple-icons/simple-icons#third-party-extensions).

See also https://github.com/simple-icons/simple-icons/issues/1382

jorgeamadosoria:
image


package-lock.json
```diff
@@ -1,8 +1,15221 @@
 {
   "name": "simple-icons-website",
   "version": "1.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
ericcornelissen: See #108

package.json
```diff
@@ -30,7 +30,8 @@
     "test:unit": "cross-env TEST_ENV=unit jest"
   },
   "dependencies": {
-    "simple-icons": "5.19.0"
+    "simple-icons": "5.19.0",
+    "sync-fetch": "0.3.1"
```
ericcornelissen: This is a `devDependency` as it's only used in the webpack config file.

public/index.pug
```diff
@@ -56,6 +79,10 @@ html(lang="en-US")
...
+        li(class="header__list-item")
+          a(class="header__list-icon popup-trigger" title="Third party extensions" href="#")
```
ericcornelissen: If it's not a link, I wouldn't use the `a` tag :thinking:

webpack.config.js
```diff
@@ -16,6 +16,27 @@ const NODE_MODULES = path.resolve(__dirname, 'node_modules');
 const OUT_DIR = path.resolve(__dirname, '_site');
 const ROOT_DIR = path.resolve(__dirname, 'public');
 
+function parseExtensions() {
+  const response = fetch(
+    'https://raw.githubusercontent.com/simple-icons/simple-icons/develop/README.md',
+  );
```
ericcornelissen: What's the motivation for not reading the README from `node_modules/simple-icons/README.md`?

public/index.pug
jorgeamadosoria: It's for consistency with the rest of the menu. Otherwise I would have to make sure this gets styled specially. 
Also, I think it fulfills it's jog as a link even if it's merely a local anchor

package.json
jorgeamadosoria: That was added automatically. I'll move it

webpack.config.js
jorgeamadosoria: That I didn't know I could 😅

package.json
ericcornelissen: You can use `--save-dev` when running `npm install` if it's a `devDependency`.

public/index.pug
ericcornelissen: Can you elaborate, the styling should come from the `.header__list-icon` class, not the `a` tag

package.json
jorgeamadosoria: there is no need if I'm going to parse from node_modules. I'm pretty sure I can remove it altogether.

public/index.pug
jorgeamadosoria: I could use a span with the class, and it should work in the same way, but it would look out of place in the menu. It's not a hard reason, it's more about me liking the menu as it is. For example, I was thinking on making the menu a list and render it with an each block in pug, because they were all so uniform. I don't know, I like it this way. It's not a technical reason.

webpack.config.js
jorgeamadosoria: Done! Also, removed the sync-fetch dependency. It turned out to be much better than before.

public/index.pug
ericcornelissen: You can do that, but then you would have to exclude this item. It's not a link, so you shouldn't use an `<a>` tag. Sure, it doesn't matter visually, but you should use the correct tag so the browser and other software is better able to understand the structure of the document.

And a `<span>` is **not** the right element, it should be a `<button>` because, well, it's a button

package-lock.json
ericcornelissen: Do you need help with this?

jorgeamadosoria: the PR does not compile if I don't upload package-lock.json

public/index.pug
jorgeamadosoria: PDF and SVG downloads are <a>, though. They behave as buttons due to a class, and the ARIA is handled with a role attribute.
I think that's better.

jorgeamadosoria: I've added role="button" according to what's said here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

It also says that the <a> tags with role="button" should handle keydown and the Space key, but that's beyond this PR since we would have to do that for all the <a role="button"> tags, not just this one.

package-lock.json
jorgeamadosoria: 108 is not merged. If merged, will it do away with this issue?

public/index.pug
ericcornelissen: > PDF and SVG downloads are , though. They behave as buttons due to a class

As far as I know downloads in HTML should be `<a>`, if that's not the case we should consider changing the element we use for the downloads.



> I've added role="button" according to what's said here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role
> 
> It also says that the tags with role="button" should handle keydown and the Space key, but that's beyond this PR since we would have to do that for all the tags, not just this one.

`role="button"` should be used when you can't use an actual button.

Why can't we use a `<button>` here? What's the motivation for using `<a>`? Is it [still](https://github.com/simple-icons/simple-icons-website/pull/127#discussion_r735145971) just consistency? If that is more important to you than semantically correct HTML than feel free to leave it as is. (In which case please don't resolve this conversation and keep it open for visibility - we can agree to disagree, but that doesn't mean to conversation is resolved).

package-lock.json
ericcornelissen: > 108 is not merged. If merged, will it do away with this issue?

#108 is on the topic of the version of the lockfile. If merged in it's current state it won't allow you to use a version of npm that forces `"lockfileVersion": 2`.

> the PR does not compile if I don't upload package-lock.json

I don't know what you mean by that, I don't see any reason why it won't work if you revert the lockfile back to version 1 (which you do not achieve by just changing this value). To revert `package-lock.json` back to version one you have to run `npm install` with an npm version below 7.0.0 (or 8.1.0 with `--lockfile-version=1`).

jorgeamadosoria: It meqns that if i don't upload a package lock json, it doesn't work on the github workflow.

What's the problem with version 2, anyways? Why stick on a partocular npm version onstead of moving to a newer one?

ericcornelissen: > It meqns that if i don't upload a package lock json, it doesn't work on the github workflow.

You do indeed have to keep the `package-lock.json` file. You could just copy the one from `master` if you didn't change `package.json` in this PR.


> What's the problem with version 2, anyways? Why stick on a partocular npm version onstead of moving to a newer one?

It's not set in stone but see #108, https://github.com/simple-icons/simple-icons/pull/6179, and other threads linked in those PRs for more context. In short, happy to change it, but not in this PR.

jorgeamadosoria: This PR changed it by installing the fetch modules. Now that I removed those maybe I can copy from master

jorgeamadosoria:
Closing in favor of #134, which is the same thing but cleaned up.

PR #131, opened 2021-10-27
Update README.md

jorgeamadosoria:
Updated the icon count to more than 2k


ericcornelissen:
Perhaps it's easier to just omit the number from the README in this repository? It doesn't seem particularly useful here and keeping it can just get outdated. What do you think @jorgeamadosoria?

jorgeamadosoria:
I agree. The website does not need to track this abd it will be easier to maintain

PR #132, opened 2021-10-29
Adding query parameters for ordering and color scheme

### LINKED ISSUE #45: Support more query parameters

---

### Kind of issue 

- [x] Feature
- [ ] Bug
- [ ] Design
- [ ] Other, namely:

### Description

As per [this comment](https://github.com/simple-icons/simple-icons/discussions/4865#discussioncomment-352284) add more query parameters than just the one for the search parameter. E.g.:

- `?colorscheme=(dark|light|system)`: This loads the website with a specific color scheme. As the color scheme is also stored in [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) I don't think this should be added to the URL automatically. Instead it may be added to link new users to the website in a specific color scheme. For existing users - those with a preference stored - the color scheme in the URL should be ignored.
- `?ordering=(alpha|color|relevance)`: This loads the website with a specific ordering of the icons. As the ordering is also stored in [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) I don't think this should b added to the URL automatically. Instead, it may be added to link new users to the website with a specific ordering. FOr existing users - those with a preference stored - the ordering in the URL should be ignored.

jorgeamadosoria:
Added request parameters for ordering (o) and color scheme (c).
Required a heavy refactoring of init methods and the corresponding test suites.
I also refactored the parameter parsing functions to use URLSearchParams and URL classes. It makes the code clearer and safer, and they are VanillaJS so no added size to be concerned about.

I've also included an initControlButton function to reduce code duplication across the system. Further refactorings can be applied to reduce the code duplication and size, but those are out of the scope of this ticket.

### DIFF
--- a/public/scripts/color-scheme.js
+++ b/public/scripts/color-scheme.js
@@ -1,4 +1,10 @@
 import { STORAGE_KEY_COLOR_SCHEME } from './storage.js';
+import {
+  paramFromURL,
+  COLOR_PARAMETER,
+  setParameterInURL,
+  initControlButton,
+} from './utils.js';
 
 const COLOR_SCHEME_DARK = 'dark';
 const COLOR_SCHEME_LIGHT = 'light';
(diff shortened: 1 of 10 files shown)


ericcornelissen:
What is your motivation for adding these as query parameters? What function does it serve?

At least for me, I wouldn't want these things to be controlled by the query parameters. I wouldn't want these things to change if someone with different parameters sends me link for simple-icons.

jorgeamadosoria:
> What is your motivation for adding these as query parameters? What function does it serve?
> 
> At least for me, I wouldn't want these things to be controlled by the query parameters. I wouldn't want these things to change if someone with different parameters sends me link for simple-icons.

didn't you open this ticket: https://github.com/simple-icons/simple-icons-website/issues/45 ?
I'm simply addressing it.

ericcornelissen:
My apologies, forgot about that ticket. However, as per #45 it's not _my_ idea but rather a suggestion from https://github.com/simple-icons/simple-icons/discussions/4865#discussioncomment-352284.

That being said, could you explain in text how your implementation is intended to work? What is the behavior of these query parameters in various scenarios?

jorgeamadosoria:
> My apologies, forgot about that ticket. However, as per #45 it's not _my_ idea but rather a suggestion from [simple-icons/simple-icons#4865 (comment)](https://github.com/simple-icons/simple-icons/discussions/4865#discussioncomment-352284).
> 
> That being said, could you explain in text how your implementation is intended to work? What is the behavior of these query parameters in various scenarios?

The params in the url would override (but NOT overwrite) localStorage. That means that the url can force the UI to ignore the local preferences, but shouldn't change the values stored.
The Url could contain up to three parameters:
- q: the existing query, no changes.
- o: the order parameter, one of order-alphabetically, order-color or order-relevance. 
- c: the color scheme parameter, one of light, dark or system.

This can be provided in any combination.

Additionally, there are a few refactorings done to remove code duplication, which I find useful to handle the parameters

Having said that, I'll be completely fine with removing the whole thing if you think we don't really need it. More than completely fine, actually. I do like the idea of having parameters, but I don't think it's all that useful and I like simple code far more than scatter shot features.

If I were to add a parameter, it would be an icon parameter for the icon detail view, nothing else.

jorgeamadosoria:
> > My apologies, forgot about that ticket. However, as per #45 it's not _my_ idea but rather a suggestion from [simple-icons/simple-icons#4865 (comment)](https://github.com/simple-icons/simple-icons/discussions/4865#discussioncomment-352284).
> > That being said, could you explain in text how your implementation is intended to work? What is the behavior of these query parameters in various scenarios?
> 
> The params in the url would override (but NOT overwrite) localStorage. That means that the url can force the UI to ignore the local preferences, but shouldn't change the values stored. The Url could contain up to three parameters:
> 
> * q: the existing query, no changes in behavior, but the parsing is refactored.
> * o: the order parameter, one of order-alphabetically, order-color or order-relevance.
> * c: the color scheme parameter, one of light, dark or system.
> 
> This can be provided in any combination.
> 
> Additionally, there are a few refactorings done to remove code duplication, which I find useful to handle the parameters
> 
> Having said that, I'll be completely fine with removing the whole thing if you think we don't really need it. More than completely fine, actually. I do like the idea of having parameters, but I don't think it's all that useful and I like simple code far more than scatter shot features.
> 
> If I were to add a parameter, it would be an icon parameter for the icon detail view, nothing else.

ericcornelissen:
> The params in the url would override localStorage.

As per my comments on this topics earlier. I don't support it implemented in that way. I don't see any reason why the URL - which someone else may be providing me with - would **override** my personal preferences. I don't want someone to send me a `simpleicons.org` link that has dark mode enabled when I prefer to use it in light mode and have declared that preference when I visited the website last time, there's simply no reason.

I'm not against adding these query parameters altogether. See https://github.com/simple-icons/simple-icons/discussions/4865#discussioncomment-352640 for more details.


> Having said that, I'll be completely fine with removing the whole thing if you think we don't really need it. More than completely fine, actually. I do like the idea of having parameters, but I don't think it's all that useful and I like simple code far more than scatter shot features.

In that case I'm unsure why you decided to implement it at all, but feel free to close this and #45 if you feel that it is an unnecessary feature. Please do leave a comment with your reasoning when closing #45.

jorgeamadosoria:
Notice that it overrides your preference, but it doesn't overwrite it.
What that means is that if you remove the query params you will again see the page as you have declared in your preferences.

This could be useful if someone wants to call your attention on how a particular set of preferences look like, or when sending a link to a new user that has no preferences set. I still don't think that's important enough, but it could be used that way.

I implemented it because it's a ticket that exists. That you have reservations about it didn't mean that you didn't want it implemented. 

Having discussed it, I have no problems diacarding this and closing the ticket as not really useful, with your input on the matter.

ericcornelissen:
> Notice that it overrides your preference, but it doesn't overwrite it.
What that means is that if you remove the query params you will again see the page as you have declared in your preferences.

I notice that and I understand the difference. I don't want to have to inspect a link, and possibly remove some of it, before I click on it to have a what I would say is normal experience.


> This could be useful if someone wants to call your attention on how a particular set of preferences look like, or when sending a link to a new user that has no preferences set. I still don't think that's important enough, but it could be used that way.

It can indeed be useful for new users, but as an existing user I wouldn't want to be bothered by this.

---

I won't approve this Pull Request in it's current state, if you can find support for it among other maintainers then feel free to move ahead with this :slightly_smiling_face:

jorgeamadosoria:
> > Notice that it overrides your preference, but it doesn't overwrite it.
> > What that means is that if you remove the query params you will again see the page as you have declared in your preferences.
> 
> I notice that and I understand the difference. I don't want to have to inspect a link, and possibly remove some of it, before I click on it to have a what I would say is normal experience.
> 
> > This could be useful if someone wants to call your attention on how a particular set of preferences look like, or when sending a link to a new user that has no preferences set. I still don't think that's important enough, but it could be used that way.
> 
> It can indeed be useful for new users, but as an existing user I wouldn't want to be bothered by this.
> 
> I won't approve this Pull Request in it's current state, if you can find support for it among other maintainers then feel free to move ahead with this 🙂

I don't care about this PR. I'll close it along with the issue. That's good enough for me.

PR #133, opened 2021-10-30
Shadow luminosity refactoring

jorgeamadosoria:
changed the value of --dm-color-shadow to allow for it to be seen better in dark color scheme.

also, added a small margin to menu items to make them show properly when selected via tabbing.

### DIFF
--- a/public/stylesheet.css
+++ b/public/stylesheet.css
@@ -41,7 +41,7 @@
   --dm-color-button-text-hover: rgb(0, 0, 0);
   --dm-color-disabled: rgb(170, 170, 170);
   --dm-color-grid-item-divider: rgb(64, 64, 64);
-  --dm-color-shadow: rgba(75, 75, 75, 0.95);
+  --dm-color-shadow: rgba(215, 215, 215, 0.95);
   --dm-color-link: rgb(0, 100, 238);
   --dm-color-link-hover: rgb(0, 72, 238);
   --dm-color-link-visited: rgb(169, 144, 189);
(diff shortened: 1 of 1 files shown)


ericcornelissen (approved):
> changed the value of `--dm-color-shadow` to allow for it to be seen better in dark color scheme.

This change looks good to me, thanks for fixing :+1: 

> also, added a small margin to menu items to make them show properly when selected via tabbing.

Can you elaborate on this change? While I'm okay with the change, I'm not seeing how this "makes them show properly when selected via tabbing" :confused:

jorgeamadosoria:
In chrome, there is a thin blue outline around the menu items that shows when tabbing only.
It doesn't affect functionally but it looks as if the menu is "crowded"

ericcornelissen:
> In chrome, there is a thin blue outline around the menu items that shows when tabbing only.
It doesn't affect functionally but it looks as if the menu is "crowded"

I see, you could alternatively use `outline: none` as is done on other elements. (I hadn't noticed the default outline on these elements before as the one in Firefox is much less pronounced)

jorgeamadosoria:
> > In chrome, there is a thin blue outline around the menu items that shows when tabbing only.
> > It doesn't affect functionally but it looks as if the menu is "crowded"
> 
> I see, you could alternatively use `outline: none` as is done on other elements. (I hadn't noticed the default outline on these elements before as the one in Firefox is much less pronounced)

changed margin for outline:none.

public/stylesheet.css
```diff
@@ -229,7 +229,7 @@ a {
   }
 }
 @media (max-width: 640px) {
-  .header {
+  x .header {
```
ericcornelissen, line 232: This doens't look right @jorgeamadosoria

jorgeamadosoria: fixed in #135

PR #137, opened 2021-11-03
Add Windows support for `clean` script

mondeja:
The same change [has been included](https://github.com/simple-icons/simple-icons/pull/6450) in simple-icons and simple-icons-pdf repositories. Note that _rimraf_ is already a dependency of _@jest/core_, so this doesn't increments `node_modules` size.


jorgeamadosoria (approved):
excellent

PR #140, opened 2021-11-03
Prevent focus on hidden copy `input`

mondeja:
There is a DOM `input` node that can be focused using <kbd>TAB</kbd> key. I've found this bug using the latest lighthouse report and reproduced locally. Adding `tabindex="-1"` resolves it.


jorgeamadosoria (approved):
sure

PR #141, opened 2021-11-03
Prevent focus on `body`

mondeja:
The `body` DOM node can be focused using <kbd>TAB</kbd> key, but this is not desirable.


jorgeamadosoria (approved):
sure.

PR #145, opened 2021-11-14
Dom manipulation to sort icon cards

### LINKED ISSUE #86: Tabbing order confusion

jorgeamadosoria:
according to #86 , CSS ordering is to be substituted with JS sorting for colors to allow tabbing to work.
This PR includes DOM sorting for all criteria, related refactorings and unit and e2e tests for this new approach.
DOM rearranging is slower than the previous solution, but that was expected and is the only solution for the tabbing problem (without updating tabIndex everywhere).


ericcornelissen (commented):
Please resolve the merge conflicts and undo unwanted changes (e.g. `package*.json`) first :slightly_smiling_face:

mondeja:
Would you mind to solve the conflicts @jorgeamadosoria?

jorgeamadosoria:
> Would you mind to solve the conflicts @jorgeamadosoria?

conflicts are solved.

mondeja:
New conflict in `public/scripts/index.js` 🔧

mondeja (commented):
New conflict in `public/scripts/index.js` 🔧

jorgeamadosoria:
> New conflict in `public/scripts/index.js` 🔧

this is done.

public/stylesheet.css
```diff
@@ -282,6 +284,29 @@ a {
   padding: 0.2rem;
 }
 
+/* Extensions */
```
mondeja: Why this is being added again? Please, rebase properly.

mondeja (dismissed):

public/index.pug
```diff
@@ -37,12 +37,12 @@ html(lang='en-US')
...
           )
 
-  body.no-js.order-alphabetically.download-svg(tabindex='-1')
+  body.no-js.order-alpha.download-svg(tabindex='-1')
     .popup_modal.popup__top.hidden
       .popup-body 
         table.extensions__table
           tr
-            td Extension
+            td Extensions
```
mondeja: This renaming has no sense. The column of the table should be singular.

public/index.pug
```diff
@@ -304,10 +307,10 @@ html(lang='en-US')
           src='//cdn.carbonads.com/carbon.js?serve=CKYIPK7M&placement=simpleiconsorg',
           type='text/javascript'
         )
-
         each icon in icons
           li.grid-item(
-            style=`--order-color: ${icon.indexByColor}`,
+            order-color=`${icon.indexByColor}`,
+            order-alpha=`${icon.indexByAlpha}`,
             data-brand=`${icon.normalizedName}`
```
mondeja: These values of attributes can be defined in pug using the variables directly:

```suggestion
            order-color=icon.indexByColor,
            order-alpha=icon.indexByAlpha,
            data-brand=icon.normalizedName
```

tests/ordering.test.js
```diff
@@ -121,30 +122,30 @@ describe('Ordering', () => {
...
 
-    initOrdering(document, localStorage);
+    initOrdering(document, localStorage, domUtils);
     expect(localStorage.hasItem).toHaveBeenCalledWith(STORAGE_KEY_ORDERING);
     expect(localStorage.getItem).toHaveBeenCalledWith(STORAGE_KEY_ORDERING);
-    expect(localStorage.setItem).not.toHaveBeenCalled();
+    expect(localStorage.setItem).toHaveBeenCalledTimes(1);
   });
 
   it('uses the stored value "color"', () => {
```
mondeja:
```suggestion
  it('uses the stored value "order-color"', () => {
```

public/stylesheet.css
```diff
@@ -536,7 +538,7 @@ input[type='search']::-ms-reveal {
 }
 
 /* Control - order */
-body.order-alphabetically #order-alpha,
+body.order-alpha #order-alpha,
 body.order-by-color #order-color,
 body.order-by-relevance #order-relevance {
```
mondeja: These selectors must be updated as well.

```suggestion
body.order-color #order-color,
body.order-relevance #order-relevance {
```

tests/dom-utils.test.js
```diff
@@ -2,14 +2,33 @@ const {
...
 } = require('../public/scripts/dom-utils.js');
 
+const orderedData = [
+  { 'order-alpha': 1, getAttribute: (order) => 1 },
+  { 'order-alpha': 2, getAttribute: (order) => 2 },
+  { 'order-alpha': 3, getAttribute: (order) => 3 },
+];
+const testData = [
+  { 'order-alpha': 3, getAttribute: (order) => 3 },
+  { 'order-alpha': 2, getAttribute: (order) => 2 },
```
mondeja: To avoid a false negative in the test if the implementation reverse ordering you could invert the order of these two items:

```suggestion
  { 'order-alpha': 2, getAttribute: (order) => 2 },
  { 'order-alpha': 3, getAttribute: (order) => 3 },
```

mondeja (dismissed):

public/index.pug
jorgeamadosoria: as you wish.

public/scripts/dom-utils.js
```diff
@@ -19,6 +19,16 @@ export function toggleVisibleElement($el) {
...
 
+export function sortChildren($el, attribute) {
+  [...$el.children]
+    .sort(
+      (a, b) =>
+        parseInt(a.getAttribute(attribute)) -
+        parseInt(b.getAttribute(attribute)),
+    )
+    .forEach((node) => $el.appendChild(node));
+}
```
mondeja, line 30: Here is a improved async version of this function to do the ordering instantly:

```suggestion
  const sorted = [...$el.children]
    .sort(
      (a, b) =>
      parseInt(a.getAttribute(attribute)) -
      parseInt(b.getAttribute(attribute)),
    )
  $el.innerHTML = ""
  sorted.slice(0, 15).forEach((node) => $el.appendChild(node))
  setTimeout(async () => {
    sorted.slice(15).forEach((node) => $el.appendChild(node));
  }, 5);
```

Resetting the HTML content and appending synchronously only the first 15 icons do the trick. I suppose that this number can be increased without decreasing performance.

public/scripts/ordering.js
```diff
@@ -1,44 +1,42 @@
...
   let preferredOrdering = DEFAULT_ORDERING;
 
   const $body = document.querySelector('body');
-  const $orderAlphabetically = document.getElementById('order-alpha');
-  const $orderByColor = document.getElementById('order-color');
-  const $orderByRelevance = document.getElementById('order-relevance');
+  const $orderAlpha = document.getElementById('order-alpha');
+  const $orderColor = document.getElementById('order-color');
+  const $orderRelevance = document.getElementById('order-relevance');
+  const $gridIcons = document.querySelector('ul.grid');
```
mondeja: This variable is not used:

```suggestion
```

mondeja (changes requested):

public/scripts/dom-utils.js
mondeja: Here is a improved async version of this function to do the ordering "instantly". Really doesn't does the ordering instantly, but seems like it because the icons placed at the top are reordered and added first to the grid:

```suggestion
export async function sortChildren($el, attribute) {
  const sorted = [...$el.children]
    .sort(
      (a, b) =>
      parseInt(a.getAttribute(attribute)) -
      parseInt(b.getAttribute(attribute)),
    )
  const firstChildren = sorted.slice(0, 15);
  $el.innerHTML = "";
  firstChildren.forEach((node) => $el.appendChild(node));
  setTimeout(async () => {
    sorted.slice(15).forEach((node) => $el.appendChild(node));
  }, 1);
}
```

Resetting the HTML content and appending first only the first 15 icons do the trick. I suppose that this number can be increased without decreasing performance.

mondeja (changes requested):

public/scripts/ordering.js
```diff
@@ -1,44 +1,42 @@
...
-  $orderByColor.disabled = false;
-  $orderByRelevance.disabled = false;
+  $orderAlpha.disabled = false;
+  $orderColor.disabled = false;
+  $orderRelevance.disabled = false;
 
   if (storage.hasItem(STORAGE_KEY_ORDERING)) {
     const storedOrdering = storage.getItem(STORAGE_KEY_ORDERING);
     selectOrdering(storedOrdering);
+    domUtils.sortChildren(document.querySelector('ul.grid'), storedOrdering);
```
mondeja: This line is not required, the ordering is executed two times in this case:

```suggestion
```

mondeja (changes requested):

public/scripts/dom-utils.js
jorgeamadosoria: Isn't setTimeout going to create a new anon function per 15 elements, thus eating memory?
What happens if you sort after having scrolled down a bit (thus the elements you see are not the first 15)?

mondeja: > Isn't setTimeout going to create a new anon function per 15 elements, thus eating memory?

Not sure if I follow you, `slice(15)` means from 15 to the end. `setTimeout` is executed one time. 

> What happens if you sort after having scrolled down a bit (thus the elements you see are not the first 15)?

That is the problem. The page scrolls to the top or the bottom when the ordering is changed. Not sure if could be worth to store the scroll height pixel before sorting and reset the same scroll position after. Something like:

```javascript
const scrollY = window.scrollY;
// after ordering
window.scrollTo(0, scrollY);
```

This can be improved computing what cards should be added first depending on the position of the screen.

jorgeamadosoria:
I can't answer on the previous thread, don't know why. 

I thought setTimeout was making more than one call. If it is like you said, then nevermind, this is fine.

I have a different issue:

This is also not going to be enough for the compact layout. Maybe instead of 15 the slice should be 30 or even 50 (compact can have up to 9 cards per line, so 3 lines would be 27 instead of 15).
Maybe that also takes care of most of the cases if people have done a little bit of scrolling, not too much.

More importantly: visually, there is not that big of a delay between ordering the whole list and ordering it by slices, so I don't know if the gain is enough to justify the complexity. This is just my opinion, however: if it is up to me, simpler is better than faster, unless the code is too slow for confort. I don't think the sorting is too slow for confort just yet.

In short, this is clever, but I don't think there is any need for it. If you still think we need it, I would recommend a larger slice to account for compact (30 cards should be fine), a comment along the lines of "Above the Fold item sorting" or something like that on the first call, and "rendering the rest of the sorted items" in the setTimeout. Something to mitigate the complexity.

Other than that, the code works just fine.

mondeja:
I think that the async implementation is necessary because the blocking would be a great performance regression.

After reflecting on it a bit, seems that the scroll before sorting should be setted to 0 in axis Y, so the page would be seen at the start. Currently, the scroll behaves erratically when sorting because travels to some random point of the page, so it is an improvement anyways.

```javascript
    $body.classList.add(selected);
    window.scrollTo(0, 0);
    domUtils.sortChildren(document.querySelector('ul.grid'), selected);
```

With this line and the asynchronous implementation seems to me that it should be fine, unless I'm missing something happening when sorting by relevance on search.

jorgeamadosoria:
> I think that the async implementation is necessary because the blocking would be a great performance regression.
> 
> After reflecting on it a bit, seems that the scroll before sorting should be setted to 0 in axis Y, so the page would be seen at the start. Currently, the scroll behaves erratically when sorting because travels to some random point of the page, so it is an improvement anyways.
> 
> ```js
>     $body.classList.add(selected);
>     window.scrollTo(0, 0);
>     domUtils.sortChildren(document.querySelector('ul.grid'), selected);
> ```
> 
> With this line and the asynchronous implementation seems to me that it should be fine, unless I'm missing something happening when sorting by relevance on search.

Scrolling to the beginning is definitely good, I agree.
I would make a domUtils function for setting the scrolling view value. Or you can leave it like that and I will abstract that later on. 

If you are going to keep this, then merge it at will.

mondeja:
Some improvements in the tests and this will be ready 👍🏼

mondeja (approved):
LGTM

PR #147, opened 2021-11-20
Fix unminified build on development mode

developStorm:
Should use `argv.mode` instead of `env.mode`

https://webpack.js.org/configuration/mode/


mondeja (approved):
Thanks!

PR #155, opened 2021-12-05
Pin to NodeJS v12 in "Automated Releases" CI workflow

mondeja:
Automated Releases workflow [is failing](https://github.com/simple-icons/simple-icons-website/actions/runs/1542135316) because default NodeJS version in Github Actions has been updated to v16 this week, see https://github.com/actions/virtual-environments/issues/4446, so pin to 12 until April, 2022, when we should update all our CI workflows in all repos to v14 or v16.


service-paradis (approved):
LGTM, thank you @mondeja!

PR #157, opened 2021-12-06
Set default download type if not setted in local storage

### LINKED ISSUE #156: Icons can't be downloaded because they have `.null` extension in href
This is happenning in production. I can't replicate it in development.

### Kind of issue 

- [ ] Feature
- [x] Bug
- [ ] Design
- [ ] Other, namely:

#### Firefox

image

#### Brave browser (Chromium based)

image

mondeja:
Fixes #156


jorgeamadosoria (approved):
this should take care of the .null error in the href reported in the bug #156

PR #159, opened 2021-12-12
Added primary changes to add page for Removed icons

### LINKED ISSUE #148: Add "museum" page for discontinued brands
As time passes, some brands disappear and are removed from the library. This represents a backwards compatibility problem with users that have these icons but keep their dependencies updated.

This is only an issue for brands or icons that are removed due to falling popularity or outright disappearing. Not every removal qualifies for this, for example:
- Brands that are added by mistake or whose guidelines change after being added in such a manner as to preclude them from being used (Fedora icon, for example). In these cases a backwards break is acceptable since it would force the user to forgo a trademark violation.
- 
- Brands that are updated due to changing design (Google trademarks changes fairly frequently). The updated design would substitute the old one, which is a minor issue, only affecting the display of historically significant icon lists, which is not the goal of this library anyways.
- Changes in name or slug. This will introduce a backwards compatibility issue, but since the icon is not changed nor removed, the icon won't qualify to be conserved: it's still live in the main repo.

A potential solution for this is a list page with the removed icons, similar to the existing main page but without the features for sorting or copying, etc. A minimal page would just list all removed SVGs so the user can potentially download it and use it manually or at most offer a permanent link to the latest version in which the icon can be found.

The page would have to be manually updated on each release to add the icons removed, of course.

jorgeamadosoria:
Added Removed icons page, to contain all deleted icons in previous version with a format similar to the one at the main page. Some icons that were removed because the brand does not allow modifications are therefore listed, but not shown in compliance with the reason they were removed (i.e., Fedora, Yandex, etc).

Pending changes: navigation to and fro the main page.

Alternatively, we can just forgo navigation and link this from the Readme in the main repository, as a note.


LitoMore:
`.DS_Store` should not be committed.

LitoMore (changes requested):
`.DS_Store` should not be committed.

sachinraja:
Just a small tip - you should [add it to your global gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c)

.gitignore
```diff
@@ -9,3 +9,4 @@ node_modules/
 .cache/
 _site/
 tests/_artifacts/
+.DS_Store
```
LitoMore: Let's do this with another PR.

LitoMore (changes requested):

.gitignore
jorgeamadosoria: Why? This is where the problem was raised, this is as good a place as any to add a file to gitignore.
It doesn't warrant a separate PR, since it is not a controversial nor complex change.
Atomicity is good, but there is no need to take it to extremes.

LitoMore: That's because your commit message `Added primary changes to add page for Removed icons` will show behind the `.gitignore` for a long time. And it doesn't seem to be related to the `.gitignore` file. It would be better if it shows `Ignore .DS_Store` for `.gitignore`.

CleanShot 2021-12-14 at 11 02 23@2x

jorgeamadosoria: > That's because your commit message `Added primary changes to add page for Removed icons` will show behind the `.gitignore` for a long time. And it doesn't seem to be related to the `.gitignore` file. It would be better if it shows `Ignore .DS_Store` for `.gitignore`.
> 
> CleanShot 2021-12-14 at 11 02 23@2x

ah, I understand. OK, that makes sense.

scripts/removed-icons.js
```diff
@@ -0,0 +1,314 @@
+module.exports = [
```
mondeja, line 1: We need a way to update this file automatically. Do you have something in mind about it @jorgeamadosoria?

public/removed.pug
```diff
@@ -0,0 +1,67 @@
...
+    link(rel='icon', type='image/x-icon', href=`${pageUrl}/images/favicon.ico`)
+    link(rel='icon', type='image/png', href=`${pageUrl}/images/favicon.png`)
+    link(rel='apple-touch-icon', href=`${pageUrl}/images/apple-touch-icon.png`)
+    link(rel='mask-icon', href=`${pageUrl}/images/logo.svg`, color='#111111')
+    link(
+      rel='stylesheet',
+      href='https://fonts.googleapis.com/css2?family=Open+Sans:wght@600&family=Roboto+Mono:wght@400;600&display=swap'
+    )
+    link(rel='license', href='license.txt')
+    link(rel='canonical', href=pageUrl)
```
mondeja, line 27: The head and the footer should be reused between the two pages. See [Template Inheritance with Pug](https://pugjs.org/language/inheritance.html).

webpack.config.js
```diff
@@ -140,6 +141,34 @@ module.exports = (env, argv) => {
...
+                versionIssue: icon.versionIssue,
+                path: icon.path,
+                title: icon.title,
+              };
+            }),
+          iconCount: removedIcons.length,
+          pageTitle: 'Removed Icons',
+          pageDescription:
+            "Removed icons in all previous versions. Some icons can't be shown since they were removed because the brand does not allow them to be modified to fit this library.",
+          pageUrl: 'https://simpleicons.org/removed',
```
mondeja, line 170: A lot of configuration can also be reused between the two pages, this should help us maintaining the project.

mondeja (changes requested):

scripts/removed-icons.js
jorgeamadosoria: No. I don't really mind updating this manually. We don't remove that many icons, this is easy to keep updated just by reading the Release notifications and getting the removed issues from there.
I'm fine with this setup.
Of course, if you can think of a way to automate it, I'm all ears. The less work, the better.

webpack.config.js
jorgeamadosoria: agreed.

scripts/removed-icons.js
mondeja: I think that it should be inside the "Auto Release" workflow, after [the step that bumps simple-icons version](https://github.com/simple-icons/simple-icons-website/blob/45979f1e6fe1489e6d94d7c6b36cc2b06f627cec/.github/workflows/auto-release.yml#L27-L31).

Technically would be "if there is a major version change, compare the icons in latest release with the new ones, and add those deleted to the file".

> No. I don't really mind updating this manually.

If so, you maybe need to create a release checklist and declare some compromise to maintain it. Anyways, please, note that we don't want to depend on a single contributor for periodical manual tasks. If you left the organization tomorrow for whatever reason someone else would be obliged to keep this, so I am totally against not automating it.

jorgeamadosoria: yeah, no I get that. Not sure how to add this to the Auto Release workflow. I'll take a look.

jorgeamadosoria:
The changes necessary to automate the removed icons  page are not worth the trouble. I'm closing this in favor of greater focus on the other open PRs on the website.

PR #173, opened 2022-01-16
Optimize `normalizeSearchTerm` function

mondeja:
Here are the performance tests done to check this optimization:

<details>
  <summary>previous.js</summary>

```javascript
const normalizeSearchTerm = function (value) {
  return value
    .toLowerCase()
    .replace(/đ/g, 'd')
    .replace(/ħ/g, 'h')
    .replace(/ı/g, 'i')
    .replace(/ĸ/g, 'k')
    .replace(/ŀ/g, 'l')
    .replace(/ł/g, 'l')
    .replace(/ß/g, 'ss')
    .replace(/ŧ/g, 't')
    .normalize('NFD')
    .replace(/[\u0300-\u036f]/g, '');
};

for (let r = 0; r < process.argv[2]; r++) {
  normalizeSearchTerm('đħıĸŀłłĸħßŧŧŧŧabcdefghi');
}
```

</details>

```sh
time node previous.js 2000000

real	0m6,801s
user	0m7,359s
sys	0m0,113s
```

<details>
  <summary>new.js</summary>

```javascript
const replacements = {
  đ: 'd',
  ħ: 'h',
  ı: 'i',
  ĸ: 'k',
  ŀ: 'l',
  ł: 'l',
  ß: 'ss',
  ŧ: 't',
};

const regex = RegExp(Object.keys(replacements).join('|'), 'g');
const regex2 = RegExp('[\u0300-\u036f]', 'g');

const normalizeSearchTerm = function (value) {
  return value
    .replace(regex, (char) => replacements[char])
    .normalize('NFD')
    .replace(regex2, '');
};

for (let r = 0; r < process.argv[2]; r++) {
  normalizeSearchTerm('đħıĸŀłłĸħßŧŧŧŧabcdefghi');
}
```

</details>

```sh
time node new.js 2000000

real	0m4,377s
user	0m4,508s
sys	0m0,017s
```

It seems that [fast-fuzzy is calling `toLocaleLowerCase()` function](https://github.com/EthanRutherford/fast-fuzzy/blob/fe56314465d057630de8ca8c102d2a9ca4aca397/src/fuzzy.js#L29) to normalize strings to search, so converting to lower case in `search.js` is not needed.


public/scripts/search.js
```diff
@@ -92,7 +92,6 @@ export default function initSearch(history, document, ordering, domUtils) {
     const result = searcher.search(query);
     let noResults = true;
     $icons.forEach(($icon) => {
-      const brandName = $icon.getAttribute('data-brand');
```
mondeja, line 95: This variable is not used

sachinraja (approved):
Maybe we should do the same for the main repo: https://github.com/simple-icons/simple-icons/blob/develop/scripts/utils.js#L26-L41

PR #174, opened 2022-01-16
Reuse copy buttons path

mondeja:
Reduces HTML file size from 2.37 MB to 1.66 MB.

### DIFF
--- a/public/index.pug
+++ b/public/index.pug
@@ -30,6 +30,11 @@ html(lang='en-US')
     )
     link(rel='license', href='license.txt')
     link(rel='canonical', href=pageUrl)
+    svg(xmlns='http://www.w3.org/2000/svg', style='display: none')
+      symbol#cp
+        path(
+          d='M11.2 0a.8.8 0 0 0-.8.8v11.4L7.26 9.44a.803.803 0 0 0-1.13.074l-1.05 1.2a.8.8 0 0 0 .073 1.13l6.33 5.54a.795.795 0 0 0 1.05 0l6.32-5.54a.8.8 0 0 0 .074-1.13l-1.05-1.2a.804.804 0 0 0-1.13-.074l-3.14 2.76V.8a.8.8 0 0 0-.8-.8zm-8 20.8a.8.8 0 0 0-.8.8v1.6a.8.8 0 0 0 .8.8h17.6a.8.8 0 0 0 .8-.8v-1.6a.8.8 0 0 0-.8-.8z'
+        )
 
   body.no-js.order-alphabetically.download-svg(tabindex='-1')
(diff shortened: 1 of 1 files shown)


sachinraja:
Not sure which browsers we're targeting but [`xlink:href` is deprecated](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use#attributes)

mondeja:
Great, it seems to work with `href`, so 6 characters less for each icon.

sachinraja:
I'm not very familiar with it, but I think we should put the symbol between [`defs`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/defs) instead of using `display: none` on the svg.

mondeja:
> I'm not very familiar with it, but I think we should put the symbol between [`defs`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/defs) instead of using `display: none` on the svg.

I've tested it and doesn't works, it adds a huge empty space at the top of the page. Seems that the dimensions must be defined.

This works:

```pug
svg(xmlns='http://www.w3.org/2000/svg', width="0", height="0")
  defs
    symbol#cp
      path(d='...')
```

And this:

```pug
svg(xmlns='http://www.w3.org/2000/svg', width="0", height="0")
  symbol#cp
    path(d='...')
```

But not this:

```pug
svg(xmlns='http://www.w3.org/2000/svg')
  defs
    symbol#cp
      path(d='...')
```

I'm fine with either way. Chosen this approach because `width="0", height="0"` is two characters longer than `style="display:none"` and seems more descriptive of what means. Added `defs` in https://github.com/simple-icons/simple-icons-website/pull/174/commits/ed99f3a494c271f74101d46298072ba7dbeaaa9b because [SVG spec v2 recommends it](https://svgwg.org/svg2-draft/struct.html#DefsElement) for accesibility reasons.

sachinraja (approved):

public/index.pug
```diff
@@ -30,6 +30,12 @@ html(lang='en-US')
     )
     link(rel='license', href='license.txt')
     link(rel='canonical', href=pageUrl)
+    svg(xmlns='http://www.w3.org/2000/svg')
```
sachinraja, line 33: Sorry, may have missed something in your comment. Should this have the `width="0"` and `height="0"` attributes?

mondeja: The SVG is hidden with a combination of `width="0"` and `height="0"` or just using `style="display:none"`, so it is the same. The SVG spec doesn't recommends any choice.

mondeja: Ups, seems that I've forgotten the `style="display:none"`. Fixed by #177

PR #175, opened 2022-01-16
Website initialization performance optimizations

mondeja:
- 10x performance optimization for `initCopyButtons` function reusing copy events functions declaration. Avoids to declare one function for each icon.
- Initialize modules in parallel splitting them.

### DIFF
--- a/public/scripts/copy.js
+++ b/public/scripts/copy.js
@@ -6,40 +6,6 @@ function setCopied($el) {
 }
 
 export default function initCopyButtons(document, navigator, fetch) {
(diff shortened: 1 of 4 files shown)


jorgeamadosoria:
Good refactoring on the copy buttons, there are a lot of those so naming the function should allow for reusability and less memory. Maybe that's a thinking that could help elsewhere too.

I don't see a huge win in async init for the modules, but OK, it doesn't hurt. 

let's merge it.

jorgeamadosoria (approved):
LGTM

PR #182, opened 2022-02-01
Remove `data-` attrs from cards and detach searcher initialization

mondeja:
Remove `data-icon` and `data-brand` attributes from icon cards. This reduces the built HTML size to 1.50 MB without affecting initialization performance.

### Summary of the changes:

- Remove  `data-icon` and `data-brand` attributes from icon cards (see [performance tests comment below](https://github.com/simple-icons/simple-icons-website/pull/182#issuecomment-1072311834)).
- Fix error when backing to order by relevance after search and changing to other order (not properly fixed by #194)
- Detach `search` module initialization to avoid blocking the main thread (great performance improvment).
- Move `options.keySelector` function outside of `Searcher.options` object to avoid lazy evaluation of the anonymous function when searching.
- Don't normalize `rawQuery` on search because is not needed.


sachinraja (commented):
Could we dynamically generate `href` based off of `data-brand` instead of `data-brand` off `href`?

mondeja:
Yes, the problem would be that we need to load all onclick events for `a.grid-item__button` and prevent the default behaviour in them, which doesn't reduces file size anyways.

sachinraja:
> which doesn't reduces file size anyways

But the code for that would only be in one place, in the JS file. Not duplicated for each icon.

jorgeamadosoria:
> > which doesn't reduces file size anyways
> 
> But the code for that would only be in one place, in the JS file. Not duplicated for each icon.

we could put the code for that in the same place where we are extracting the href to slug now.
My concern is that for this we would need to keep the data-icon and then add the href via the JS, so in effect we would end up with the same amount of HTML.
@mondeja's solucion is fine, imo.

jorgeamadosoria:
@mondeja the code is alright. I'm always in favor of removing as much code as possible.

My only concern is that by reducing the HTML size we do improve the loading time of the page, but we need to increase the amount of JS that needs to be executed. Did you run any tests to see how much performance gains we have, really?

jorgeamadosoria (approved):
No comments. Only concern is about actual performance gains (reducing HTML is good, but it increases JS code necessary for processing)

mondeja:
Thanks for the review. I intend to merge this after #194, but definitely will double check the possible performance decrease 👍🏼

mondeja:
What I can see after some performance tests:

#### Loading the page

- `new Searcher` initialization takes ~60ms more time
- HTML parsing and all other rendering actions (interesections computing, HTML parsing, painting...) takes less time, about ~150ms between all
- The space in memory needed to render the whole page is decreased about a 10%

#### Searching

Placing the [`keySelector` function outside the `options` object](https://github.com/simple-icons/simple-icons-website/pull/182/commits/30643e67bf010904aa87768a3d6912cc778d767b), search performance is even slightly improved because the options [are passed to `searchCore` in `fast-fuzzy`](https://github.com/EthanRutherford/fast-fuzzy/blob/fe56314465d057630de8ca8c102d2a9ca4aca397/src/fuzzy.js#L495) so the function is lazily evaluated if is anonymously defined.

---

After this, I consider that this is an improvement in performance, although being a decreasement in `Search` initialization, the other performance improvements make it worth it.

---

EDIT: note that the detaching of the searcher initialization implemented in https://github.com/simple-icons/simple-icons-website/pull/182/commits/8243f05d0cbe10655c32f8dc686149ea176bd099 (after this comment) has solved this possible problem because the searcher initialization doesn't blocks the loading anymore.

mondeja:
> But the code for that would only be in one place, in the JS file. Not duplicated for each icon.

I'm seeing now this comment... not sure if would worth it, but of course feel free to try it 👍🏼

PR #196, opened 2022-03-18
Add structured data

### LINKED ISSUE #87: Add structured data
### Kind of issue 

- [x] Feature
- [ ] Bug
- [ ] Design
- [ ] Other, namely:

### Description

Add [structured data](https://developers.google.com/search/docs/guides/intro-structured-data) to the website. This could make the results in e.g. Google a bit nice when searching for "Simple Icons".

Any opinions on whether we should add this (@simple-icons/maintainers)?

mondeja:
Closes #87

Hi @simple-icons/maintainers! It's time to add structured data to the website for optimize search results in engines like Google. The data will be in the format specified by @adamrusted  in https://github.com/simple-icons/simple-icons-website/issues/87#issuecomment-873632286. Is generated making calls to the Github API. I need your approval because your names will be used in the metadata of the page. Of course, if you don't want, we can add a file with usernames to ignore.

If you've more ideas about [other fields to add to the organization](https://schema.org/Organization), please post them here. Cheers!

#### Current data for each maintainer

```json
{
  "@type": "Person",
  "name": "Your full name (extracted from your Github account)",
  "jobTitle": "Maintainer",
  "url": "The URL of your Github account (maybe you want your website?)",
  "image": "https://the-url/of/your/github/avatar (like https://avatars.githubusercontent.com/u/23049315?v=4)"
}
```


dirien (approved):
LGTM

service-paradis (approved):

runxel (approved):

webpack.config.js
```diff
@@ -55,7 +58,84 @@ if (process.env.TEST_ENV) {
...
 
-module.exports = (env, argv) => {
+const pageDescription = `${icons.length} Free SVG icons for popular brands.`,
+  pageTitle = 'Simple Icons',
+  pageUrl = 'https://simpleicons.org',
+  logoUrl = 'https://simpleicons.org/icons/simpleicons.svg';
+
+async function generateStructuredData() {
+  const getSimpleIconsMembers = () => {
+    return new Promise(async (resolve, reject) => {
```
sachinraja: I think we should be able to use async/await directly instead of wrapping it in Promises

jorgeamadosoria:
Why add the names of the maintainers into the structured data? what is gained by doing that?

webpack.config.js
```diff
@@ -55,7 +58,82 @@ if (process.env.TEST_ENV) {
...
+    );
+    if (fs.existsSync(siMembersCacheFilePath)) {
+      try {
+        const siMembersFileContent = fs.readFileSync(siMembersCacheFilePath, {
+          encoding: 'UTF8',
+        });
+        resolve(JSON.parse(siMembersFileContent));
+      } catch (error) {
+        reject(error);
+      }
```
sachinraja:
```js
const getSimpleIconsMembers = async () => {
  const siMembersCacheFilePath = path.join(
    os.tmpdir(),
    'simple-icons-members.json'
  );
  if (fs.existsSync(siMembersCacheFilePath)) {
    const siMembersFileContent = fs.readFileSync(siMembersCacheFilePath, {
      encoding: 'UTF8',
    });
    return JSON.parse(siMembersFileContent);
  } else {
    // ...
  }
};
```
@mondeja sorry, should have been more clear. I meant we can do something like the following instead of using all the promise wrappers.

mondeja:
> Why add the names of the maintainers into the structured data? what is gained by doing that?

I'm not sure. Just the search engines would show better results, I guess. You can read this documentation https://developers.google.com/search/docs/advanced/structured-data/intro-structured-data

If you don't want to add your name we can tweak the implementation to avoid adding some. Personally, I haven't problems about showing my name as is published in Github.

webpack.config.js
mondeja: And what about the other `resolve`? Is inside a `.then` block because `Promise.all` is being used.

Feel free to push the change that you are thinking directly in the PR 👍🏼

jorgeamadosoria:
> > Why add the names of the maintainers into the structured data? what is gained by doing that?
> 
> I'm not sure. Just the search engines would show better results, I guess. You can read this documentation https://developers.google.com/search/docs/advanced/structured-data/intro-structured-data
> 
> If you don't want to add your name we can tweak the implementation to avoid adding some. Personally, I haven't problems about showing my name as is published in Github.

no, it's not a personal matter. I ask because you seem to have to add a full file just to GET the names from the github api. Kind of involved for something that is there just to be there, you know?

webpack.config.js
sachinraja: pushed the changes

sachinraja (approved):

sachinraja:
> I ask because you seem to have to add a full file just to GET the names from the github api. Kind of involved for something that is there just to be there, you know?

That file is not necessary, it's only for caching so you don't have to query the GitHub API multiple times if you rebuild. Either way that info will be in the built HTML file.

jorgeamadosoria:
> > I ask because you seem to have to add a full file just to GET the names from the github api. Kind of involved for something that is there just to be there, you know?
> 
> That file is not necessary, it's only for caching so you don't have to query the GitHub API multiple times if you rebuild. Either way that info will be in the built HTML file.

Necessary or not, it's there and that's its only function. So I have to weigh the code being added against the benefit.... and  I don't know if listing maintainers is worth adding so much code.

There has to be other more useful things to add to the data. like a search field to appear in the google results, for example.

mondeja:
> a search field to appear in the google results

Added in https://github.com/simple-icons/simple-icons-website/pull/196/commits/5139e913d5b5dc056c78e6e630a138ebacb11480

Issue #214, opened 2022-09-22
URL not indexed on Bing, Yahoo and DuckDuckGo


alanjacobmathew:
### Kind of issue 

- [ ] Feature
- [ ] Bug
- [ ] Design
- [x] Other, namely:

### Description

-  Simple icons website(simpleicons.org) are not indexed by  Bing & Yahoo and hence not indexed on DuckDuckGo , although teh github pages are indexed.
- `http://simpleicons.p2hp.com/` Not sure what this page is but even the favicon is of simpleicons.org

 

![image](https://user-images.githubusercontent.com/33965848/191743382-89b75df7-fa71-4c55-816c-562ed3bc75e7.png)
![image](https://user-images.githubusercontent.com/33965848/191743893-29237038-1151-4c20-8f61-1245f1ee59f3.png)

mondeja:
### Solutions

- [x] **Bing**: Add `https://simpleicons.org` to their index [here](https://www.bing.com/webmasters/submiturl?siteUrl=https://simpleicons.org). Also, we must add a meta tag and a sitemap XML file to the web and add a CNAME record to the DNS server. 
- [x] **Yandex**: add `https://simpleicons.org` to their index [here](https://webmaster.yandex.com/sites/add/). The steps are almost the same.

PR #216, opened 2022-09-22
Add sitemap

### LINKED ISSUE #215: Add sitemap
We've just one page but seems that some search engines index better sites with sitemap. Most likely related to #214

mondeja:
This PR adds a *sitemap.xml* file to the website. Resolves #215, contributes to #214.

The SVG files have been formatted because I've added the XML plugin for Prettier.


LitoMore:
LGTM. We could submit it to see how it works.

PR #218, opened 2022-09-25
Add localization

mondeja:
Adds a simple localization system for the website using multiple bundles along with the Spanish and Fench translations.

https://user-images.githubusercontent.com/23049315/192127494-d62c9c18-0a80-4549-81ee-0fa2eafdafd0.mp4


scripts/i18n.js
```diff
@@ -0,0 +1,146 @@
...
+
+export const updateTranslations = async () => {
+  const index = await fs.readFile(INDEX_PATH, 'utf8');
+  const msgids = extractTranslatableStringsFromIndex(pugLex(index));
+  const currentIso = new Date().toISOString();
+
+  for (const lang of LANGUAGES) {
+    let po;
+    const poPath = path.join(ROOT_DIR, 'locales', `${lang}.po`);
+    if (!fsSync.existsSync(poPath)) {
```
LitoMore: `fs.exist()` has been deprecated. Let's use `fs.stat()` or `fs.access()` instead.

But:

> Using fs.stat() to check for the existence of a file before calling fs.open(), fs.readFile(), or fs.writeFile() is not recommended. Instead, user code should open/read/write the file directly and handle the error raised if the file is not available.

So we can just use a `try-catch` instead of the `if (fs.exist(path)) else`.

Reference: https://nodejs.org/dist/latest/docs/api/fs.html#fsstatpath-options-callback

scripts/i18n.js
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env node
+/**
+ * @fileoverview
+ * Updates the translations of the project, located in the
+ * directory locales/
+ */
+
+import fs from 'node:fs/promises';
+import fsSync from 'node:fs';
```
LitoMore: Let's change all `fs` usages to `node:fs/promises`.

Our `webpack.config.js` is using ESM as well. So feel free to to top-level async/await.

mondeja:
I'm missing a section in the contributing guide about how to add new locales.

mondeja:
Now the implementation looks good, merging 🐳

PR #225, opened 2022-09-26
Translate icons in website using `loc` aliases

### LINKED ISSUE #222: Show localized brand names using `loc` aliases
Use `aliases.loc` if available to show the localized name of a brand in a locale page. 

Take into account that we're using universal language codes with two letters like `es`, `fr`... ([ISO 639-1](https://en.wikipedia.org/wiki/ISO_639-1)) but the aliases could be in a `es_ES` or `fr-FR` format ([IETF language tag](https://en.wikipedia.org/wiki/IETF_language_tag)).

mondeja:
Resolves #222

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -236,7 +236,6 @@ export default async (env, argv) => {
         iconsDataBySlugs[icon.slug],
         languages,
       ),
-      aliases: icon.aliases,
     };
   });
 
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -13,6 +13,7 @@ import alphaSort from './scripts/alpha-sorting.js';
 import colorSort from './scripts/color-sorting.js';
 import GET from './scripts/GET.js';
 import {
+  DEFAULT_LANGUAGE,
   getLanguages,
   loadTranslations,
   updateTranslations,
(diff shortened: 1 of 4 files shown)


webpack.config.js
```diff
@@ -290,40 +294,58 @@ export default async (env, argv) => {
...
-      ),
+      ...languages.map((lang) => {
+        // Add localized title for the icons in the property `localizedTitle`
+        const currentLangIcons =
+          lang === DEFAULT_LANGUAGE
+            ? icons
+            : icons.map((icon) => {
+                if (icon.aliases && icon.aliases.loc) {
+                  for (const locale of Object.keys(icon.aliases.loc)) {
+                    if (locale.substring(0, lang.legth) === lang) {
```
LitoMore: Some brands may have multiple translations like this:

```json
"loc": {
  "en-GB": "Colour",
  "en-SG": "ColorSG",
  "zh-CN": "颜色",
  "zh-TW": "色彩",
  "zh-HK": "顏色",
}
```

Do we need to force the file name to ISO Language Code + ISO Country Code `/^[a-z]{2}-[A-Z]{2}$/`? Like this: https://www.fincher.org/Utilities/CountryLanguageList.shtml

mondeja: > Do we need to force the file name to ISO Language Code + ISO Country Code /^[a-z]-[A-Z]$/? Like this: https://www.fincher.org/Utilities/CountryLanguageList.shtml

I think that is not necessary, I'll tweak the implementation 🔧

PR #252, opened 2022-09-30
Remove `pre-push` hook

LitoMore:
~~The `build` and `test:unit` take a very long time at the `pre-push` state. It's not necessary to make the production build and unit test before every push.~~

Remove the whole `pre-push` hook.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname $0)/_/husky.sh"
-
-npm run lint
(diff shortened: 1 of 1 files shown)


.husky/pre-push
```diff
@@ -2,5 +2,3 @@
 . "$(dirname $0)/_/husky.sh"
 
 npm run lint
```
mondeja: I think that we must just remove this file, the linting is done before commits

mondeja (changes requested):

mondeja (approved):

PR #253, opened 2022-09-30
Fix icon colors

LitoMore:
### Bug preview

CleanShot 2022-09-30 at 23 34 21@2x

### How to reproduce

1. Open https://simpleicons.org
2. Change your system appearance settings to light mode
3. Change page color scheme to `system color scheme`
4. Open the detail view for an icon


mondeja (approved):
Thanks 🥇

PR #256, opened 2022-10-02
Adjust `icon request` and `report oudated` buttons

LitoMore:
We received some icon update requests without any description since we added a report below the icon:

- https://github.com/simple-icons/simple-icons/issues/7901
- https://github.com/simple-icons/simple-icons/issues/7879
- https://github.com/simple-icons/simple-icons/issues/7880
- https://github.com/simple-icons/simple-icons/issues/7909

Our previous `report outdated icon` link is:

```
https://github.com/simple-icons/simple-icons/issues/new?labels=icon+outdated&template=icon_update.md&title=Update%20${icon.title}%20icon
```

Opening this link does not use any issue template. This is why their description is empty.

BTW, we may not need to add a report button to every icon. Adding a feature not commonly used in such an important position can be cluttered. So let's remove it.

Users usually find our contact information or feedback link at the very bottom of the page. So I moved the report feature to the below of `Submit a request`.

CleanShot 2022-10-02 at 17 09 35@2x

CleanShot 2022-10-02 at 17 04 32@2x

For the `Submit a request`, previously we would only show this link when the search results were empty. But in fact, we are using a fuzzy search library for filtering icons. Like the pictures on the right side.

In this case, we should show them the `Icon missing?`.

I updated those links to issue template links. Now users will see some required forms when creating the issue.

### DIFF
--- a/tests/e2e.test.js
+++ b/tests/e2e.test.js
@@ -159,19 +159,14 @@ describe('Search', () => {
     },
   );
 
-  it('always show report buttons', async () => {
+  it('show report buttons when icons are missing on search', async () => {
     const $iconMissing = await page.$('#icon-missing');
     const $iconOutdated = await page.$('#icon-outdated');
     expect(await isVisible($iconMissing)).toBeTruthy();
     expect(await isVisible($iconOutdated)).toBeTruthy();
 
     const $searchInput = await page.$('#search-input');
-    await $searchInput.type(' ');
-    expect(await isVisible($iconMissing)).toBeTruthy();
-    expect(await isVisible($iconOutdated)).toBeTruthy();
-
-    await page.$eval('#search-input', (el) => (el.value = ''));
-    await $searchInput.type('litomore');
+    await $searchInput.type('this is definitely not going to match');
     expect(await isVisible($iconMissing)).toBeTruthy();
     expect(await isVisible($iconOutdated)).toBeTruthy();
   });
@@ -277,6 +272,13 @@ describe('Search', () => {
 
     const $orderRelevance = await page.$('#order-relevance');
     expect(await isVisible($orderRelevance)).toBeTruthy();
+  });
+
+  it('shows the "no results" message if no brand was found', async () => {
+    await page.type('#search-input', 'this is definitely not going to match');
+    await page.screenshot({
+      path: path.resolve(ARTIFACTS_DIR, 'desktop_no-search-results.png'),
+    });
   });
 });
 
--- a/tests/e2e.test.js
+++ b/tests/e2e.test.js
@@ -159,6 +159,23 @@ describe('Search', () => {
     },
   );
 
+  it('always show report buttons', async () => {
+    const $iconMissing = await page.$('#icon-missing');
+    const $iconOutdated = await page.$('#icon-outdated');
+    expect(await isVisible($iconMissing)).toBeTruthy();
+    expect(await isVisible($iconOutdated)).toBeTruthy();
+
+    const $searchInput = await page.$('#search-input');
+    await $searchInput.type(' ');
+    expect(await isVisible($iconMissing)).toBeTruthy();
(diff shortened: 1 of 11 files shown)


tests/e2e.test.js
```diff
@@ -180,11 +169,6 @@ describe('Search', () => {
     expect(await isHidden($searchClear)).toBeTruthy();
   });
 
-  it('does not show the "no results" message on load', async () => {
```
mondeja, line 183: Would you mind to update these tests rather than removing them?

mondeja (changes requested):

tests/e2e.test.js
LitoMore: Updated.

tests/e2e.test.js
```diff
@@ -170,6 +159,23 @@ describe('Search', () => {
...
+    expect(await isVisible($iconMissing)).toBeTruthy();
+    expect(await isVisible($iconOutdated)).toBeTruthy();
+
+    const $searchInput = await page.$('#search-input');
+    await $searchInput.type(' ');
+    expect(await isVisible($iconMissing)).toBeTruthy();
+    expect(await isVisible($iconOutdated)).toBeTruthy();
+
+    await page.$eval('#search-input', (el) => (el.value = ''));
+    await $searchInput.type('litomore');
```
mondeja: Please, rewrite this as the old test, "litomore" is too much generic and doing fuzzy searching could lead to the test fail in the future. Better "this is definitely not going to match" or something like that.

tests/e2e.test.js
```diff
@@ -282,27 +278,6 @@ describe('Search', () => {
     const $orderRelevance = await page.$('#order-relevance');
     expect(await isVisible($orderRelevance)).toBeTruthy();
   });
-
-  it('shows the "no results" message if no brand was found', async () => {
-    await page.type('#search-input', 'this is definitely not going to match');
-    await page.screenshot({
-      path: path.resolve(ARTIFACTS_DIR, 'desktop_no-search-results.png'),
-    });
```
mondeja: This screenshot is really useful reviewing the CI, please don't remove it.

tests/e2e.test.js
```diff
@@ -170,6 +159,23 @@ describe('Search', () => {
...
   );
 
+  it('always show report buttons', async () => {
+    const $iconMissing = await page.$('#icon-missing');
+    const $iconOutdated = await page.$('#icon-outdated');
+    expect(await isVisible($iconMissing)).toBeTruthy();
+    expect(await isVisible($iconOutdated)).toBeTruthy();
+
+    const $searchInput = await page.$('#search-input');
+    await $searchInput.type(' ');
```
mondeja: This test is very confusing. Really is testing that typing a space in the search bar does not find anything, but this is not clear to me reading it. In fact, do not display icons when typing only whitespaces seems like a bug because if I add other letter after them, the icons reappear...

tests/e2e.test.js
```diff
@@ -170,6 +159,23 @@ describe('Search', () => {
...
+    expect(await isVisible($iconMissing)).toBeTruthy();
+    expect(await isVisible($iconOutdated)).toBeTruthy();
+
+    await page.$eval('#search-input', (el) => (el.value = ''));
```
mondeja: This is not possible to do from the user perspective, so I think that we should not test it.

mondeja (changes requested):

tests/e2e.test.js
```diff
@@ -170,6 +159,23 @@ describe('Search', () => {
     },
   );
 
+  it('always show report buttons', async () => {
```
mondeja:
```suggestion
  it('show report buttons when icons are missing on search', async () => {
```

mondeja (changes requested):

mondeja (approved):
Thanks!

PR #271, opened 2023-01-16
Add Japanese locale

adamrusted:
VERY rough translation from English to Japanese. Could do with validation before we include it.


mondeja (approved):
Thank you @adamrusted!

PR #281, opened 2023-04-02
Add support for copying slug

LitoMore:
This is more convenient for users of `shields.io` and `simple-icons-cdn`.

This allows users to copy the slug by clicking the icon title.

### DIFF
--- a/public/index.pug
+++ b/public/index.pug
@@ -433,7 +433,9 @@ html(lang=lang)
                   rel='noopener',
                   target='_blank'
                 ) #{ icon.license.type }
-              h2.grid-item__title #{ icon.localizedTitle || icon.title }
+              h2.grid-item__title.copy-button.copy-slug(
+                data-slug=`${ icon.slug }`
+              ) #{ icon.localizedTitle || icon.title }
             .grid-item__footer
               button(
                 class=`grid-item__color copy-button copy-color ${icon.light ? "contrast-light" : "contrast-dark"}`,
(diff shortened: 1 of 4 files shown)


public/index.pug
```diff
@@ -433,7 +433,9 @@ html(lang=lang)
                   rel='noopener',
                   target='_blank'
                 ) #{ icon.license.type }
-              h2.grid-item__title #{ icon.localizedTitle || icon.title }
+              h2.grid-item__title.copy-button.copy-slug(
+                data-slug=`${ icon.slug }`
```
mondeja: We are currently extracting the slug from `href` value of the icon SVG download button to avoid `data-` attributes using the function `iconHrefToSlug`. Would you mind to get the slug using that strategy instead and remove the `data-slug` attributes?

You can see an example of this at [this line (import)](https://github.com/simple-icons/simple-icons-website/blob/6f87c65f89c34f05e4937ec9e5aa4f6f251cf080/public/scripts/download-type.js#L2) and [these lines (usage)](https://github.com/simple-icons/simple-icons-website/blob/6f87c65f89c34f05e4937ec9e5aa4f6f251cf080/public/scripts/download-type.js#L58-L59). Just go to the parent traversing the DOM and locate the `a[download]` link node.

mondeja (changes requested):
This is so great, the style is really clever and the functionality makes a lot of sense @LitoMore, thanks!

I have just one suggestion to avoid a lot of duplicated data.

public/index.pug
LitoMore: Good point. Thank you!

mondeja (approved):

Issue #284, opened 2023-04-05
Partial renderings to avoid performance problems


mondeja:
### Kind of issue 

- [x] Feature
- [ ] Bug
- [ ] Design
- [ ] Other, namely:

### Description

The problem seems to be one of the next (or all or some of them):

- [x] In some devices or browsers, `loading=lazy` attribute for lazy loading of images don't work? This problem seems solved by #291 altough IE11 is not supported as well (the same problem of loading=lazy)
- Even with lazy loading of images working, a lot of DOM nodes could potentially freeze the loading of the page.

Maybe an intersection observer or a virtual DOM from a library could solve this problem.

mondeja:
### TODO

- Investigate if [content-visibility](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility) could help here.
- Maybe a deferred renderization of entire icon grid items using the intersection observer implemented in #291 could also help as `contain-intrinsic-size` is set.

mondeja:
I suspect that `contain-intrinsic-size` is not properly set for the compact layout 🛠️  (done in #293)

kidonng:
> * In some devices or browsers, `loading=lazy` attribute for lazy loading of images don't work?

There are two issues with `loading="lazy"`. It has worse support than `IntersectionObserver` (I'm looking at you, Safari), and it is more eager than one may expect (browser typically loads content more than two viewports away).

> * This problem seems solved by [Add intersection observer for lazy loading of images #291](https://github.com/simple-icons/simple-icons-website/pull/291) altough IE11 is not supported as well (the same problem of loading=lazy)

I _feel_ the site is snappier than when I noticed this issue previously, but it is a very subjective measurement.

I don't know how many IE users the website has, but _optimizing_ for it is certainly not worth it.

mondeja:
After the intersection observer API implementation I've noted that the scroll behaves a bit erratically. This is due to the fact that `contain-intrinsic-size` is not properly setted because all icon grid items don't have the same height. Not sure how to properly solve this problem nor if it is something that we should care about as most users I suppose that are using the search bar to look for an icon.

A solution **maybe** could be to force all icon cards to be of the same height merging the metadata rows into one like in the next image, but we'll also need to add `...` after icon titles that don't fit in one line, which is probably not an option because the user will need to open the details modal to see the brand name.

![image](https://user-images.githubusercontent.com/23049315/230347828-eab16790-b88b-4917-916d-5fe3ccf7a557.png)

[...]

PR #291, opened 2023-04-05
Add intersection observer for lazy loading of images

mondeja:
Contributes to #284

I'm not sure if this solution will avoid all loading problems, but seems better than the current implementation.  Added a smooth opacity transition on image loads too.

### DIFF
--- a/public/scripts/iobserver.js
+++ b/public/scripts/iobserver.js
@@ -1,0 +1,45 @@
+/**
+ * @fileoverview
+ * Lazy loading of images using an intersection observer.
+ */
+
+export default (document) => {
+  const images = document.querySelectorAll('img[d-src]');
+
+  if ('IntersectionObserver' in window) {
+    const lazyLoad = (target) => {
+      const io = new IntersectionObserver(
+        (entries, observer) => {
+          entries.forEach((entry) => {
+            if (entry.isIntersecting) {
+              const img = entry.target;
+              const src = img.getAttribute('d-src');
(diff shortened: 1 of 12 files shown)


public/scripts/iobserver.js
```diff
@@ -0,0 +1,45 @@
...
+    // Browser does not support IntersectionObserver,
+    // so we load using the loading=lazy attribute
+    for (const img of images) {
+      const src = img.getAttribute('d-src');
+      img.removeAttribute('d-src');
+      img.onload = () => {
+        img.classList.add('ld');
+      };
+      img.setAttribute('loading', 'lazy');
+      img.setAttribute('src', src);
```
kidonng, line 42: This seems unnecessary because `loading=lazy` support implies `IntersectionObserver` support.

mondeja: Thanks, removing that on #293

mondeja: This implementation is not exhaustive, if a browser like IE11 does not support `loading=lazy` just loads the images as normal ignoring the attribute. It is only added here in the case that IE11 or other browsers starts supporting it.

kidonng: > It is only added here in the case that IE11 or other browsers starts supporting it.

[IE is deprecated](https://endoflife.date/internet-explorer), while other browsers all support it in their latest versions.

I'm just nit-picking that `'IntersectionObserver' in window` is practically always true nowadays.

PR #302, opened 2023-08-29
Update README to use CDN Image

adamrusted:


mondeja (approved):
Great, I'll rebase ASAP with the test fix and merge. Would you mind to update it in [simple-icons-font](https://github.com/simple-icons/simple-icons-font) as well @adamrusted?

adamrusted:
Done. Looking at this, is it worth adding a test to the website to ensure the Discord link could be clicked? If I'm honest, I didn't even realise we were testing for that kind of stuff!

mondeja:
> Looking at this, is it worth adding a test to the website to ensure the Discord link could be clicked? If I'm honest, I didn't even realise we were testing for that kind of stuff!

Is really overkilling. I'm just maintaining the tests as is, not updating them since long time ago. We've accepted some new features without tests.

In fact, I've rewritten the whole website in Rust, see it at https://wasm.simpleicons.org (the repo is https://github.com/mondeja/simple-icons-website-rs) to add proper tests, better performance and the rest of features that are missing here.

PR #316, opened 2023-12-06
Future-proof website for guidelines object format.

adamrusted:
This PR serves to add some logic around which item we display in the 'guidelines' field. Should mean nothing immediately breaks with the release of https://github.com/PeterShaggyNoble/simple-icons/commit/f26ed39fb482ecb739109db0c93ca5626231f6d1 - and we can adapt to remove the query whenever we decide how we want/need to display trademark notices.


PeterShaggyNoble (changes requested):
I think that if design limitations are restricting us to only be able to display one of the `guidelines` links for the foreseeable then we should default to the trademark guidelines as they generally have greater legal implications than brand guidelines, and some will often include a link to the brand guidelines. What do you reckon?

adamrusted:
Yeah, I can expand it out so that if trademark info is there, that gets shared - otherwise, we show the brand guidelines.

webpack.config.js
```diff
@@ -294,7 +294,10 @@ export default async (env, argv) => {
     const plainAliases = getIconPlainAliases(iconsDataBySlugs[icon.slug]);
 
     return {
-      guidelines: icon.guidelines,
+      guidelines:
+        typeof icon.guidelines != 'object'
```
PeterShaggyNoble: Should we not be using a "not strictly equal to" check here?
```suggestion
        typeof icon.guidelines !== 'object'
```

webpack.config.js
```diff
@@ -294,7 +294,10 @@ export default async (env, argv) => {
...
+          ? icon.guidelines
+          : icon.guidelines?.trademark ? icon.guidelines.trademark : icon.guidelines.branding,
```
PeterShaggyNoble: As this part of the code will only be executed _if_ `guidelines` is an object then there shouldn't be a need to check that the `guidelines` exist.
```suggestion
          : icon.guidelines.trademark ? icon.guidelines.trademark : icon.guidelines.branding,
```

PeterShaggyNoble (changes requested):
Not sure if either or both of these will take care of the linting issue. Execution looks good to me otherwise.

adamrusted:
@mondeja / @LitoMore - any idea what is causing the website testing issue? Have we missed something in the title checking logic?

PeterShaggyNoble:
> @mondeja / @LitoMore - any idea what is causing the website testing issue? Have we missed something in the title checking logic?

Looks like it's failing due to changing the casing of "KinoPoisk" to "Kinopoisk" in [#10084](https://github.com/simple-icons/simple-icons/pull/10084). But it's not just failing here, [it happened](https://github.com/simple-icons/simple-icons-website/actions/runs/7236414451/job/19714951683) during the build of the site on Sunday, too.

PeterShaggyNoble:
#317 should fix the issue.

PeterShaggyNoble (approved):

mondeja (approved):

PR #322, opened 2024-04-08
Use bot token for releases

LitoMore:


.github/workflows/publish.yml
```diff
@@ -37,13 +39,12 @@ jobs:
         run: npm ci
       - name: Build Website
         run: npm run build
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mondeja, line 41: The build process uses the `GITHUB_TOKEN` environment variable because needs to call the API several times. Would you mind to use the bot token here?

mondeja (changes requested):

.github/workflows/publish.yml
```diff
@@ -13,6 +13,8 @@ jobs:
         with:
           app-id: ${{ vars.BOT_APP_ID }}
           private-key: ${{ secrets.BOT_PRIVATE_KEY }}
+          owner: 'simple-icons'
+          repositories: 'simple-icons,simple-icons-website'
```
mondeja: Why this line is needed?

LitoMore: See https://github.com/actions/create-github-app-token?tab=readme-ov-file#repositories. We're accessing `simple-icons/simple-icons` on the build step and `simple-icons/simple-icons-website` on the publish step.

mondeja: But we are accessing the public part of the API for `simple-icons/simple-icons`. It is needed to explicitly define the repos in that case?

LitoMore: Not exactly. We are using this API below for read organization members:

```js
const siOrgMembers = await githubAPI.GET('/orgs/simple-icons/members');
```

But you're right. This data is used to generate SEO `application/ld+json` information. Our private members may not want to appear on the list.

LitoMore: You could ignore the failing CI, it caused by the pull request was created from my forked repository.

See https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

It's a little bit tricky. I can push the branch to this repository and create a pull request to re-trigger the CI for this pull request...

mondeja (approved):

PR #329, opened 2024-04-15
Add XO linter

### LINKED ISSUE #327: Remove dependency on `simple-icons/simple-icons-eslint`
In an effort to slim down the repositories we maintain, this is a proposal to remove the reliance on our own eslint plugin.

LitoMore:
Closes #327


public/scripts/dom-utils.js
```diff
@@ -1,49 +1,51 @@
...
+    // eslint-disable-next-line unicorn/explicit-length-check
     Math.min(nFirstChildren, newChildren.length) || newChildren.length;
   const firstChildren = newChildren.slice(0, nFirstChildren);
-  $el.innerHTML = '';
-  firstChildren.forEach((node) => $el.appendChild(node));
+  $element.innerHTML = '';
+  for (const node of firstChildren) $element.append(node);
   setTimeout(() => {
-    newChildren.slice(nFirstChildren).forEach((node) => $el.appendChild(node));
+    for (const node of newChildren.slice(nFirstChildren)) $element.append(node);
```
mondeja, line 24: Why are you using `append` here instead of `appendChild`? The method `append` also accepts strings, which could be less safer in a typed context and prone to errors when testing using mocks.

public/scripts/iobserver.js
```diff
@@ -3,43 +3,48 @@
  * Lazy loading of images using an intersection observer.
  */
 
-export default (document) => {
+export default function iObserver(document) {
```
mondeja: Just to be enough explicit, let's call this `intersectionObserver`...
```suggestion
export default function intersectionObserver(document) {
```

public/scripts/search.js
```diff
@@ -1,12 +1,12 @@
...
+import {Searcher} from 'fast-fuzzy';
+import {ORDER_RELEVANCE} from './ordering.js';
+import {debounce} from './utils.js';
 
 const QUERY_PARAMETER = 'q';
 
 const getQueryFromParameter = (location, parameter) => {
   const query = new URLSearchParams(location.search).get(parameter);
-  return query ? query : '';
+  return query ?? '';
```
mondeja: What about this?

```suggestion
  return query || '';
```

The double question mark looks really unintuitive to me comparing against a default.

public/scripts/search.js
```diff
@@ -79,13 +80,13 @@ export default (history, document, ordering, domUtils) => {
...
       const $gridChildren = document.querySelector('ul.grid').children;
-      domUtils.replaceChildren(
-        document.querySelector('ul.grid'),
-        getNonIcons($gridChildren).concat($allIcons),
-      );
-      // and reset to the preferred ordering
+      domUtils.replaceChildren(document.querySelector('ul.grid'), [
+        ...getNonIcons($gridChildren),
+        ...$allIcons,
+      ]);
```
mondeja: `concat` is faster and here performance in totally crucial, I've refactored this code several times for this reason. See https://stackoverflow.com/questions/48865710/spread-operator-vs-array-concat

public/scripts/search.js
```diff
@@ -96,11 +97,11 @@ export default (history, document, ordering, domUtils) => {
     $orderRelevance.classList.add('last__button');
     $orderColor.classList.remove('last__button');
 
-    // fuzzy search
+    // Fuzzy search
     let result = searcher.search(query);
     const $gridChildren = document.querySelector('ul.grid').children;
     const nonIcons = getNonIcons($gridChildren);
-    result = nonIcons.concat(result);
+    result = [...nonIcons, ...result];
```
mondeja: The same here

scripts/color-sorting.js
```diff
@@ -51,23 +51,23 @@ const rgbToHsl = (rgb) => {
   s = delta === 0 ? 0 : delta / (1 - Math.abs(2 * l - 1));
 
   // Multiply l and s by 100 to get the value in percent, rather than [0,1]
-  s = +(s * 100);
-  l = +(l * 100);
-  return { h, s, l };
+  s = Number(s * 100);
+  l = Number(l * 100);
```
mondeja: I don't understand this change. Is `Number` really required here?

scripts/i18n.js
```diff
@@ -51,18 +52,23 @@ export const loadTranslations = async () => {
...
     for (const item of po.items) {
       translations[item.msgid] = item.msgstr[0];
     }
+
     locales[locale] = translations;
   }
 
   const defaultTranslations = {};
   for (const msgid in locales[Object.keys(locales)[0]]) {
+    if (!msgid) continue;
```
mondeja: Empty msgids are not valid PO entries, why do you check them here?

mondeja (changes requested):
An initial review without executing the code. I do not reviewed tests nor webpack config.

public/scripts/dom-utils.js
LitoMore: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v51.0.1/docs/rules/prefer-dom-node-append.md

public/scripts/search.js
LitoMore: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v51.0.1/docs/rules/prefer-spread.md

But I can ignore the error.

scripts/color-sorting.js
LitoMore: https://eslint.org/docs/latest/rules/no-implicit-coercion

Then `+()` is not necessary here either.

scripts/i18n.js
LitoMore: https://eslint.org/docs/latest/rules/guard-for-in

You have no idea what the `locales` is. The code between `const locales = {}` and `locales[locale] = translations` may bugged. So the the `locales` is possibly `{}`.

public/scripts/search.js
mondeja: Please, ignore it. This would be a huge regression in performance.

scripts/color-sorting.js
mondeja: Yeah, you can remove the `+()` coercion, which is totally useless code when dealing with numbers.

public/scripts/color-scheme.js
```diff
@@ -9,13 +9,13 @@ const DEFAULT_COLOR_SCHEME = COLOR_SCHEME_SYSTEM;
...
 const CLASS_LIGHT_MODE = 'light';
 
-export default (document, storage) => {
+export default function colorScheme(document, storage) {
   let activeColorScheme = DEFAULT_COLOR_SCHEME;
 
   const $body = document.querySelector('body');
-  const $colorSchemeDark = document.getElementById('color-scheme-dark');
-  const $colorSchemeLight = document.getElementById('color-scheme-light');
-  const $colorSchemeSystem = document.getElementById('color-scheme-system');
```
mondeja, line 18: You've removed all usages of `document.getElementById` but the mock is not removed. Please, search for `getElementById` and remove all the mocks and the code related with them.

scripts/i18n.js
mondeja: Is better to raise an error here rather than continuing silently.

tests/assets.test.js
```diff
@@ -13,19 +13,21 @@ const extractContextsSvgPaths = (tokens) => {
...
-          // title, path
-          prevToken.val.substring(1, prevToken.val.length - 1),
-          token.val.substring(1, token.val.length - 1),
+          // Title, path
+          previousToken.val.slice(1, -1),
+          token.val.slice(1, -1),
         ];
       }
+
+      return undefined;
```
mondeja: More explicit is `return null` here.

package.json
```diff
@@ -22,8 +22,9 @@
     "build:dev": "webpack --mode development",
     "build:watch": "npm run build:dev -- --watch",
     "clean": "del-cli _site tests/_artifacts",
-    "format": "prettier --write . && eslint . --fix",
-    "lint": "prettier --check . && eslint .",
+    "prettier": "prettier --ignore-unknown \"**/*.!(js|jsx|mjs|cjs|ts|tsx|mts|cts)\"",
+    "format": "npm run prettier -- --write . && xo --fix",
+    "lint": "npm run prettier -- --check . && xo",
```
mondeja: Would you mind to separate `prettier` and `xo` commands like in https://github.com/simple-icons/simple-icons/pull/10739?

tests/dom-utils.test.js
```diff
@@ -1,129 +1,131 @@
-import { jest } from '@jest/globals';
-
+/* eslint accessor-pairs: ["error", {"setWithoutGet": false}] */
```
mondeja: I would prefer to maintain a `get` method rather to need to know what does this comment. It would be just:

```js
  get: jest.fn(() => $element.children),
```

mondeja (changes requested):
Additionally, I think that is worth to update the *.nvmrc* version to `v20` and the *Dockerfile* to use `node:20-alpine` in this PR, but as you prefer.

scripts/i18n.js
LitoMore: I just refactored the code with:

```js
const defaultTranslations = Object.keys(Object.values(locales).at(0)).map(
  (msgid) => ({
    [msgid]: msgid,
  }),
);
```

It will throw an error if the data is incorrect.

package.json
LitoMore: What do you mean? I didnt get you. I already moved the `prettier --ignore-unknown \"**/*.!(js|jsx|mjs|cjs|ts|tsx|mts|cts)\"` to the `prettier` script. This should be clear enough.

mondeja: See the second statement of the linked PR:

> Shows that `xo --fix` is being executed in format script. Currently, after Prettier is executed `xo --fix` runs but no output is shown, which seems that Prettier is stopped at some point.

Currently, after running `npm run prettier` the output looks freeze until XO ends, which looks like Prettier is stopped at some point. That's why I introduced the `xo:fix` script in the main repo.

scripts/i18n.js
mondeja: That is not the same. You're generating a `Array<Object<String, String>>`, but the others locales are defined as `Object<String, String>`. The code works because the `|| msgid` in the `i18n()` function.

All of these is not needed for the default locale. Just apply the next diff, please:

```diff
-  const defaultTranslations = Object.keys(Object.values(locales).at(0)).map(
-    (msgid) => ({
-      [msgid]: msgid,
-    }),
-  );
-
-  locales.en = defaultTranslations;
+  locales.en = {};
```

mondeja (approved):
Great! I've archived the simple-icons-eslint package.

PR #330, opened 2024-05-20
Fix website not using correct localized icon title.

kj415j45:
Live version:
image

After fix:
image

PR #331, opened 2024-05-20
Use `pull_request_target` event to run verification workflows

mondeja:
When a PR is executed openning from a fork the `BOT_APP_ID` environment variable is not set. See [this run](https://github.com/simple-icons/simple-icons-website/actions/runs/9156088123/job/25169779098?pr=330) of #330.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - master
-  pull_request:
+  pull_request_target:
 
 jobs:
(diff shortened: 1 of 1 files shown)


adamrusted (approved):

LitoMore:
Be careful with this action. See https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git

mondeja:
Okay, let's go with #332

PR #340, opened 2024-08-13
Adjusted sticky navigation CSS to prevent text overlap at the top of the header

AMatisse:
I have made a modification to the CSS for the sticky navigation. Previously, when scrolling, a portion of the text would appear at the top of the sticky header.

Live version:
image
⏬
After fix:
image

### DIFF
--- a/public/stylesheet.css
+++ b/public/stylesheet.css
@@ -364,7 +364,7 @@ a {
   flex-direction: row;
   flex-wrap: wrap-reverse;
   margin: 0 calc(-1 * var(--side-offset)) 1.2rem;
-  padding: 0.8rem var(--side-offset);
+  padding: 0.8rem var(--side-offset) 0.65rem;
   position: sticky;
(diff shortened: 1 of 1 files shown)


public/stylesheet.css
```diff
@@ -366,7 +366,7 @@ a {
   margin: 0 calc(-1 * var(--side-offset)) 1.2rem;
   padding: 0.8rem var(--side-offset);
```
mondeja: To make it pixel-perfect, could you adjust the bottom padding as well?

```suggestion
  padding: 0.8rem var(--side-offset) 0.65rem;
```

mondeja (commented):
Thanks! I just have a suggestion...

public/stylesheet.css
AMatisse: You're right, it's perfect like that
I applied the suggestion

mondeja (approved):

